PR #1, opened 2023-08-25
Refactor to prevent mod-lists from accessing FQM schema

bvsharp:
## Purpose
mod-list-app: Use the FQM API instead of going to its DB schema

Implementation of above ticket requires these changes in mod-fqm-manager.

## Testing
- [x] All unit tests passed
- [x] List refreshes and exports work with the new APIs

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -57,6 +57,16 @@
           "permissionsRequired": ["fqm.query.async.results.get"]
         },
         {
+          "methods": ["GET"],
+          "pathPattern": "/query/{query-id}/sortedIds",
+          "permissionsRequired": ["fqm.query.async.results.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/query/contents",
+          "permissionsRequired": ["fqm.query.async.results.get"]
+        },
+        {
           "methods": ["POST"],
           "pathPattern": "/query",
           "permissionsRequired": ["fqm.query.async.post"]
--- a/src/main/java/org/folio/fqm/resource/FqlQueryController.java
+++ b/src/main/java/org/folio/fqm/resource/FqlQueryController.java
@@ -2,17 +2,18 @@ package org.folio.fqm.resource;
 
 import lombok.RequiredArgsConstructor;
 import org.folio.fqm.service.QueryManagementService;
+import org.folio.querytool.domain.dto.ContentsRequest;
 import org.folio.querytool.domain.dto.QueryDetails;
 import org.folio.querytool.domain.dto.QueryIdentifier;
 import org.folio.querytool.domain.dto.ResultsetPage;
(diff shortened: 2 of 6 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -56,6 +56,16 @@
           "pathPattern": "/query/{query-id}",
           "permissionsRequired": ["fqm.query.async.results.get"]
         },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/query/{query-id}/sortedIds",
```
bvsharp, line 61: New API required by mod-lists to get sorted list of ids from mod-fqm-manager during list refresh.

src/main/java/org/folio/fqm/resource/FqlQueryController.java
```diff
@@ -34,6 +33,17 @@ public ResponseEntity<QueryDetails> getQuery(UUID queryId, Boolean includeResult
     return queryDetails.map(ResponseEntity::ok).orElseGet(() -> ResponseEntity.notFound().build());
   }
 
+  @Override
+  public ResponseEntity<List<Map<String, Object>>> getContents(UUID entityTypeId, List<String> fields, List<UUID> ids) {
```
bvsharp, line 37: Right now this is a GET request with entityTypeId, fields, and ids as request parameters. Passing the list of fields and ids as request parameters is kind of messy, so we may want to consider doing this as a POST with a request body or something similar instead.

mweaver-ebsco: I have a feeling you're right. If the list of fields or IDs is particularly large (which I'm sure it will be at some point), we'll run into the size limit on the URL in the GET. I'm kind of inclined to hold off until we know it's really a problem before we add a new endpoint, though

Fun fact: last I checked, there's no formally defined size limit for that in the HTTP RFCs, so it can (and does) change from one platform to the next 🙂

descriptors/ModuleDescriptor-template.json
mweaver-ebsco: (no changes or anything here... This just seems like a convenient place to float an idea)

I wonder how resource-intensive it'd be to move the sorting to mod-lists... Then, we'd be able to just request the data for a list by doing a query for a list of fields where `id in (id1, id2, id3, etc)` - on the one hand, it'd simplify the FQM by not needing this endpoint. It'd also relieve some strain on the DB, which is always good. On the other hand, we'd need to make sure each mod-lists instance has enough RAM to sort large lists, which could get expensive.

Another idea: add something to the query request schema to define the sort order. In that case, we could using the existing endpoints to do a query from mod-lists to FQM for a list of fields where `id in (id1, id2, id3, etc)` ordered by a, b, c. Then, we'd still be able to sort in the DB and keep the mod-lists instance size down

mweaver-ebsco (approved):

PR #6, opened 2023-08-29
modifying-entity-types: adding enrollment date and fixing defect

kjain110:
Migrating previous PR's


src/main/resources/db/changelog/changes/v1.0.0/sql/derived-tables/create-view-user-details.sql
```diff
@@ -18,6 +18,7 @@ SELECT
         UserDetails.jsonb -> 'personal' ->> 'mobilePhone' as Mobilephone,
         UserDetails.jsonb -> 'personal' ->> 'dateOfBirth' as DateOfBirth,
         UserDetails.jsonb ->> 'expirationDate'::text AS Expiration_date,
+        UserDetails.jsonb ->> 'enrollmentDate'::text AS Enrollment_date,
```
bvsharp, line 21: Not a huge deal, but should we do these aliases the same as in the other files (lowercase letters with underscores) just to keep things consistent?

bvsharp (approved):

PR #15, opened 2023-09-18
MODFQMMGR-14: Add maven failsafe integration test for Dockerfile

julianladisch:
https://issues.folio.org/browse/MODFQMMGR-14

Add a maven-failsafe-plugin integration test that runs the Dockerfile image so that we test the shaded fat uber jar.

This would have caught https://issues.folio.org/browse/MODFQMMGR-11 and https://issues.folio.org/browse/MODFQMMGR-12 early.

## Purpose
Test coverage for shaded fat uber jar and Dockerfile image.

## Approach
maven-failsafe-plugin integration test

## Learning
Always add maven failsafe integration test for any module.

### DIFF
--- a/src/test/java/org/folio/fqm/ModFqmManagerIT.java
+++ b/src/test/java/org/folio/fqm/ModFqmManagerIT.java
@@ -74,7 +74,7 @@ class ModFqmManagerIT {
   }
 
   @Test
-  public void health() {
+  void health() {
     // request without X-Okapi-Tenant
     when()
       .get("/admin/health")
@@ -84,7 +84,7 @@ class ModFqmManagerIT {
   }
 
   @Test
-  public void installAndUpgrade() {
+  void installAndUpgrade() {
     String body = """
                   {"module_to": "999999.0.0",
                    "parameters": [{"key": "loadReference", "value": "true"},


src/test/java/org/folio/fqm/ModFqmManagerIT.java
```diff
@@ -0,0 +1,127 @@
...
+                                  {"key": "loadSample",    "value": "true"} ]
+                  }
+                  """;
+    postTenant(body);
+
+    // migrate from 0.0.0, migration should be idempotent
+    body = """
+           {"module_from": "0.0.0",
+           """ + body.substring(1);
+    postTenant(body);
```
mweaver-ebsco, line 100: I just want to make sure I'm understanding this part... This is saying to upgrade from 0.0.0 to 999999.0.0, and that should be idempotent because it's already installed 999999.0.0? Is that right?

src/test/java/org/folio/fqm/ModFqmManagerIT.java
```diff
@@ -0,0 +1,127 @@
...
+
+  private void smokeTest() {
+    given()
+      .header("X-Okapi-Tenant", "beeuni")
+      .contentType("application/json")
+      .when()
+      .get("/entity-types")
+      .then()
+      .statusCode(200)
+      .body("$.size()", is(3));
```
mweaver-ebsco, line 124: The 3 seems a little fragile, since it's so specific (every time we add a new entity type, this test will need updated). Maybe `.body("$.size()", greaterThanOrEqualTo(3));`?

mweaver-ebsco (approved):
This is slooooooow on an M1 mac (the M1's ARM architecture means the x86 docker base image has to run in a fairly slow x86 emulator, so it takes over a minute for the mod-fqm-manager container to start up. I'm definitely glad the container is being put in a `static` field so that it doesn't have to start twice), but I'm okay with that if it means fewer surprise breakages like we've been seeing 😀

src/test/java/org/folio/fqm/ModFqmManagerIT.java
mweaver-ebsco: Totally makes sense. Thanks!

src/test/java/org/folio/fqm/ModFqmManagerIT.java
mweaver-ebsco: Fair enough. Eventually, we'll probably want to figure out a nicer way to get the number of entity types (maybe a `select count(*) from entity_types` directly on the container from the test or something), but I'm okay with manually updating it for now.

PR #45, opened 2023-10-12
MODFQMMGR-31:referred contact type should be drop down

kjain110:
https://issues.folio.org/browse/MODFQMMGR-31


src/main/resources/db/changelog/changes/v1.0.0/sql/entity-type-definitions/insert-user-details-definition.sql
```diff
@@ -154,7 +154,30 @@ INSERT INTO entity_type_definition (id, derived_table_name, definition)
...
+                          {
+                            "value": "phone",
+                            "label": "Phone"
+                          },
+                          {
+                            "value": "mobile",
+                            "label": "Mobile"
+                          }
+
+                      ]
```
kjain110, line 180: Need to add the drop down values like this since for preferred contact type does not come from db, they are hard coded values in the users-module

mweaver-ebsco: At least it's easy

mweaver-ebsco (approved):

PR #51, opened 2023-10-24
MODFQMMGR-57 Fix f_unaccent() call to use index

mweaver-ebsco:
This commit updates the drv_item_callnumber_location view to use the mod_inventory_storage version of f_unaccent(), so that it can make better use of indexes. In order to make this change, the commit also moves that view definition from a SQL file to XML, so that we can easily access the tenant name. The view was reformatting as part of that, but the only substantial change is the different schema for f_unaccent()


src/main/resources/db/changelog/changes/v1.0.0/drv_item_callnumber_location_create.xml
```diff
@@ -0,0 +1,78 @@
...
+        src_inventory_item.jsonb ->> 'itemLevelCallNumber'::text                                      AS item_level_call_number,
+        src_inventory_item.jsonb ->> 'itemLevelCallNumberTypeId'::text                                AS item_level_call_number_typeid,
+        call_item_number_type_ref_data.item_level_call_number_type_name,
+        (src_inventory_item.jsonb -> 'effectiveCallNumberComponents'::text) ->>'callNumber'::text     AS item_effective_call_number,
+        (src_inventory_item.jsonb -> 'effectiveCallNumberComponents'::text) ->>'typeId'::text         AS item_effective_call_number_typeId,
+        call_number_type_ref_data.item_effective_call_number_type_name,
+        loclib_ref_data.id                                                                            AS item_effective_library_id,
+        loclib_ref_data.item_effective_library_name,
+        loclib_ref_data.item_effective_library_code,
+        "left"(lower(${tenant_id}_mod_inventory_storage.f_unaccent((src_inventory_item.jsonb -> 'status'::text) ->> 'name'::text)), 600) AS item_status,
```
mweaver-ebsco, line 31: This is the only real change in this PR. I reformatted the view when moving it to XML, but didn't really change anything other than adding the `${tenant_id}_mod_inventory_storage.` to the `f_unaccent()` call here.

mweaver-ebsco: Technically, the preconditions are different, too. I stripped those down to just the views actually used by drv_item_callnumber_location

bvsharp (approved):

PR #59, opened 2023-10-25
Rework item views to use instance metadata indexes

bvsharp:
## Purpose
[MODFQMMGR-51](https://issues.folio.org/browse/MODFQMMGR-51): queries using instance_updated_date and instance_created_date take a very long time 

## Approach
Rework item views to use instance metadata indexes for faster querying


ncovercash (approved):
LGTM

mweaver-ebsco (approved):

PR #63, opened 2023-10-27
MODFQMMGR-67 Enable batched inserts

mweaver-ebsco:
This just re-introduces the changes that were accidentally merged before they'd be reviewed

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -17,6 +17,8 @@ spring:
     hikari:
       # time in ms to wait for db connection
       initialization-fail-timeout: 1800000
+      data-source-properties:
+        reWriteBatchedInserts: true
   sql:
     init:
       # to boot up application despite of any DB connection issues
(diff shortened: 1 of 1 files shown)


ncovercash (approved):
hitting the right button this time, 🙃

src/main/resources/application.yml
```diff
@@ -31,6 +33,8 @@ spring:
       hibernate:
         dialect: org.hibernate.dialect.PostgreSQLDialect
         format_sql: true
+        jdbc:
+          batch_size: 50
```
mweaver-ebsco, line 37: I don't think this will actually do anything, since (afaik) we don't actually use hibernate in mod-fqm-manager. But I figured if we have some hibernate properties at all, we should also have this one :P

src/main/resources/application.yml
```diff
@@ -17,6 +17,8 @@ spring:
     hikari:
       # time in ms to wait for db connection
       initialization-fail-timeout: 1800000
+      data-source-properties:
+        reWriteBatchedInserts: true
```
mweaver-ebsco, line 21: This one should make a difference, even without hibernate. It makes it so that DB doesn't internally break batched inserts back up into single inserts when running.

PR #67, opened 2023-11-02
[MODFQMMGR-72] Remove instance_title_searchable field

ncovercash:
## Purpose
`instance_title_searchable` was causing more problems than it was worth, so we removed it.

## Approach
Add new Liquibase changelogs to recreate the view and update the associated entity_type_definition.

## Learning
- Turns out Postgres [will not allow you to remove columns when running a `CREATE OR REPLACE VIEW` query](https://stackoverflow.com/questions/65117930/postgresql-error-cannot-drop-columns-from-view), so we have to drop it first.  We can assume it's already there, though, since previous version's changes will be created before this changelog is executed.

## Bonus
- ignore `.env` files

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.0.2/update_item_holdingsrecord_entity_type_definition.xml
+++ b/src/main/resources/db/changelog/changes/v1.0.2/update_item_holdingsrecord_entity_type_definition.xml
@@ -1,0 +1,201 @@
+<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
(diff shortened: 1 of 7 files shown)


kjain110 (approved):

bvsharp (changes requested):
The way things are set up right now, we have 2 different derived views with the instance_title_searchable field: drv_item_details and drv_item_holdingsrecord_instance. This will need to be removed from the view and entity type definition for both.

src/main/resources/db/changelog/changes/v1.0.2/update_item_holdingsrecord_entity_type_definition.xml
```diff
@@ -0,0 +1,201 @@
...
+            {
+              "name": "instance_title",
+              "dataType": {
+                "dataType": "stringType"
+              },
+              "labelAlias": "Instance title",
+              "visibleByDefault": true
+            },
+            {
+              "name": "instance_title_searchable",
```
bvsharp, line 55: This entry should be removed

ncovercash: Doh, oops

bvsharp (approved):

PR #72, opened 2023-11-07
MODFQMMGR-76: Periodically refresh materialized views

bvsharp:
## Purpose
Refresh materialized views to address [MODFQMMGR-76](https://issues.folio.org/browse/MODFQMMGR-76)

## Testing
- [x] All unit tests passed
- [x] Tenant can be onboarded with no errors
- [x] Materialized views are refreshed periodically


src/test/java/org/folio/fqm/repository/MaterializedViewRefreshRepositoryTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+class MaterializedViewRefreshRepositoryTest {
```
bvsharp, line 13: I figured that since the logic for this is so simple, we really only need to ensure that the SQL that gets executed is what we expect, and thus could just use Mockito for this (rather than the docker test-db stuff).

descriptors/ModuleDescriptor-template.json
```diff
@@ -144,6 +155,12 @@
       "description": "Run a query synchronously and get results",
       "visible":  true
     },
+    {
+      "permissionName": "fqm.materializedViews.post",
```
bvsharp, line 159: Note: this permission (along with fqm.query.purge) is not part of fqm.query.all

ncovercash (approved):

kjain110 (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -91,6 +96,12 @@
           "pathPattern": "/query/purge",
           "unit": "hour",
           "delay": "1"
+        },
+        {
+          "methods": [ "POST" ],
```
kjain110, line 101: small thing, we can remove the spaces before and after opening and ending of ""

descriptors/ModuleDescriptor-template.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "@artifactId@-@version@",
-  "name": "The module descriptor for mod-fqm-manager.",
+  "name": "FQM Manager Module",
```
mweaver-ebsco, line 3: ❤️ 
This drives me nuts every time I look at the list of installed modules in any environment 🙂

src/main/resources/swagger.api/mod-fqm-manager.yaml
```diff
@@ -44,6 +44,19 @@ paths:
...
+      tags:
+        - materializedViews
+      description: Refresh all materialized views for a tenant.
+      responses:
+        '204':
+          description: 'Views refreshed'
+        '400':
+          $ref: '#/components/responses/badRequestResponse'
+        '500':
+          $ref: '#/components/responses/internalServerErrorResponse'
```
mweaver-ebsco, line 59: Thanks for putting this here :) I meant to say something about it definitely not going in folio-query-tool-metadata (since that would put it in the edge module, too), and totally forgot

PR #77, opened 2023-11-20
[MODFQMMGR-87] Move FQM-specific code from lib-fqm-query-processor

ncovercash:
# [Jira MODFQMMGR-87](https://issues.folio.org/browse/MODFQMMGR-87)

We are moving the code removed from https://github.com/folio-org/lib-fqm-query-processor/pull/13 into this repository (essentially the entire `org.folio.fqm.lib` package, with the exception of `FqlValidationService`).

The same PR also made `FqlValidationService` into a proper Spring bean, so we use that instead of the previous manual instantiation.

Most of the code was moved as-is, with the following notable exceptions:

- `MetaDataRepository` was merged into `EntityTypeRepository`
- `MetaDataService` was merged into `EntityTypeService`
- Some services and util classes that were in `org.folio.fqm.lib.repository` were moved into service and util packages, as applicable
- Added `@Service` and similar stereotype annotations to moved code
- Removed schema qualifiers from imported library repositories/services
- Split `ResultSetRepository`'s tests into two separate files (one with arrays and another without, since the existing data provider required on schema/tenant names to provide different data)
- Parameterized `FqlToSqlConverterServiceTest` ✨ 

Also:
- Brought in a few dependencies from `lib-fqm-query-processor`

# SHOULD BE MERGED AFTER https://github.com/folio-org/lib-fqm-query-processor/pull/13


pom.xml
```diff
@@ -26,9 +26,12 @@
     <folio-spring-base.version>7.1.2</folio-spring-base.version>
     <folio-query-tool-metadata.version>1.0.0</folio-query-tool-metadata.version>
     <mapstruct.version>1.5.2.Final</mapstruct.version>
-    <lib-fqm-query-processor.version>1.0.0</lib-fqm-query-processor.version>
+    <lib-fqm-query-processor.version>modfqmmgr-87-SNAPSHOT</lib-fqm-query-processor.version>
```
ncovercash, line 29: This should be `1.1.0-SNAPSHOT`, but I wanted to test against the other branch in CI.  Will update once the `lib-fqm-query-processor` PR is merged.

mweaver-ebsco (approved):

pom.xml
```diff
@@ -63,7 +66,7 @@
       <version>${folio-query-tool-metadata.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.folio</groupId>
+      <groupId>com.github.folio-org</groupId>
```
mweaver-ebsco, line 69: Should this be changed?

ncovercash: part of the above; needed to use jitpack's repository rather than the regular FOLIO one. This is just for temporary testing in CI; I'll revert it once the lib- PR is merged.

ncovercash: see https://www.jitpack.io/#folio-org/lib-fqm-query-processor/modfqmmgr-87-SNAPSHOT for more details

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceTest.java
```diff
@@ -0,0 +1,301 @@
...
+              field("field3").notEqual(true)
+            )
+          )
+      )
+    );
+  }
+
+  @ParameterizedTest
+  @MethodSource("jooqConditionsSource")
+  void shouldGetJooqConditionForFqlEqualsStringCondition(String label, String fqlCondition, Condition expectedCondition) {
```
bvsharp, line 286: Very cool! The smallest nitpick ever: The test name could probably be something like shouldGetJooqConditionForFqlCondition now.

bvsharp (approved):

PR #81, opened 2023-12-05
MODFQMMGR-79 Add support for filter value getters

mweaver-ebsco:
This commit lets us use different implementations fields in the WHERE vs SELECT clauses of the generated SQL queries that get run by FQM. This allows us to make better use of indexes without mangling the data in the returned result set


src/main/java/org/folio/fqm/utils/SqlFieldIdentificationUtils.java
```diff
@@ -21,13 +20,34 @@ public class SqlFieldIdentificationUtils {
...
-  public static Field<Object> getSqlField(EntityTypeColumn entityTypeColumn) {
-    if (entityTypeColumn.getValueGetter() == null) {
-      // No value getter is defined for this column. Use column name as the SQL field name.
+  public static Field<Object> getSqlResultsField(EntityTypeColumn entityTypeColumn) {
+    if (entityTypeColumn.getValueGetter() != null) {
+      return getMappedField(entityTypeColumn);
+    }
+    // No value getter is defined for this column. Use column name as the SQL field name.
+    return field(entityTypeColumn.getName());
+  }
```
mweaver-ebsco, line 31: I went ahead and refactored this to reorder the `return`s to better reflect how we pick what to use for a "field" when building a SQL query (so use the value getter if it exists. If not, fall back to the column name). It's kind of a pointless change in this method,  but it made the new `filter` method much clearer and I figured the 2 nearly identical methods should have a similar structure

src/test/java/org/folio/fqm/utils/SqlFieldIdentificationUtilsTest.java
```diff
@@ -10,25 +10,59 @@
...
+  }
+
+  @Test
+  void testSqlFilterFieldWhenValueGetterMissing() {
+    EntityTypeColumn entityTypeColumn = new EntityTypeColumn()
+      .name("derived_column_name_01");
+    var actualField = SqlFieldIdentificationUtils.getSqlFilterField(entityTypeColumn);
+    var expectedField = field(entityTypeColumn.getName());
+    assertEquals(expectedField, actualField);
+  }
```
mweaver-ebsco, line 57: Hopefully, Sonar doesn't get too upset that I pretty much just duplicated the 2 existing tests and changed 1 line in each...

bvsharp (approved):

ncovercash (approved):

PR #85, opened 2023-12-12
MODFQMMGR-80: Build the SELECT clause from field accessors defined in the entity type

bvsharp:
## Purpose
Build the SELECT clause from field accessors defined in the entity type. Update entity type definitions to be used with new technique.

## Testing
- [x] All unit tests passed
- [x] Tenant onboarded with no errors
- [x] Query can be run successfully and results retrieved for each entity type
- [x] UserFriendlyQueryService in mod-lists still works for each entity type
- [x] GET /entity-types/{entityTypeId}/columns/{columnName}/values still works for derived fields using materialized views

## Notes
- This ticket does not add filterValueGetters, so only valueGetters will be used initially
- For getting the filter field, filterValueGetter now takes precedence over valueGetter if both are present
- Commented out the script to create the POL entity type since it doesn't work properly right now
- Now that idValueGetter logic is implemented, we actually don't need the idView field in the entity type definitions. This field can be removed from the definition.

### DIFF
--- a/src/main/java/org/folio/fqm/repository/IdStreamer.java
+++ b/src/main/java/org/folio/fqm/repository/IdStreamer.java
@@ -1,6 +1,7 @@
 package org.folio.fqm.repository;
 
 import lombok.RequiredArgsConstructor;
+import org.folio.fqm.exception.ColumnNotFoundException;
 import org.folio.fqm.exception.EntityTypeNotFoundException;
 import org.folio.fqm.model.IdsWithCancelCallback;
 import org.folio.fqm.service.DerivedTableIdentificationService;
@@ -50,9 +51,8 @@ public class IdStreamer {
                               int batchSize,
                               Consumer<IdsWithCancelCallback> idsConsumer) {
     EntityType entityType = getEntityType(entityTypeId);
-    String derivedTable = derivedTableIdentifier.getDerivedTable(entityType, fql, sortResults);
     Condition sqlWhereClause = this.fqlToSqlConverter.getSqlCondition(fql.fqlCondition(), entityType);
-    return this.streamIdsInBatch(entityType, derivedTable, sortResults, sqlWhereClause, batchSize, idsConsumer);
+    return this.streamIdsInBatch(entityType, sortResults, sqlWhereClause, batchSize, idsConsumer);
   }
 
   /**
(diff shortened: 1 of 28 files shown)


src/main/java/org/folio/fqm/repository/IdStreamer.java
```diff
@@ -50,9 +51,8 @@ public int streamIdsInBatch(UUID entityTypeId,
                               int batchSize,
                               Consumer<IdsWithCancelCallback> idsConsumer) {
     EntityType entityType = getEntityType(entityTypeId);
-    String derivedTable = derivedTableIdentifier.getDerivedTable(entityType, fql, sortResults);
     Condition sqlWhereClause = this.fqlToSqlConverter.getSqlCondition(fql.fqlCondition(), entityType);
-    return this.streamIdsInBatch(entityType, derivedTable, sortResults, sqlWhereClause, batchSize, idsConsumer);
+    return this.streamIdsInBatch(entityType, sortResults, sqlWhereClause, batchSize, idsConsumer);
```
mweaver-ebsco, line 55: Can it be? The derived view isn't getting used anymore? 🎉

src/main/java/org/folio/fqm/repository/IdStreamer.java
```diff
@@ -90,23 +88,29 @@ public List<UUID> getSortedIds(String derivedTableName,
...
-                               String derivedTableName,
                                boolean sortResults,
                                Condition sqlWhereClause,
                                int batchSize,
                                Consumer<IdsWithCancelCallback> idsConsumer) {
     return jooqContext.transactionResult(ctx -> {
+      String idValueGetter = entityType
+        .getColumns()
+        .stream()
+        .filter(col -> col.getName().equals(ID_FIELD_NAME))
```
mweaver-ebsco, line 99: Tiniest of nitpicks:
```suggestion
        .filter(col -> ID_FIELD_NAME.equals(col.getName()))
```
That way, we're protected from a potential `NullPointerException` if `col.getName()` is `null` somehow. It also makes it so every call to `.equals()` happens in `ID_FIELD_NAME` which slightly increases the odds that the JVM can optimize it at runtime somehow

ncovercash (approved):

PR #87, opened 2023-12-18
MODFQMMGR-81 Refactor how we do case-insenstive comparisons in queries

mweaver-ebsco:


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -68,145 +72,168 @@ public Condition getSqlCondition(String fqlCriteria, EntityType entityType) {
...
    * Converts the given FQL condition to the corresponding SQL query
    */
-  public Condition getSqlCondition(FqlCondition<?> fqlCondition, EntityType entityType) {
-    return sqlConverters.getOrDefault(fqlCondition.getClass(), (c, e) -> falseCondition()).apply(fqlCondition, entityType);
+  // The unchecked warning is suppressed because sqlConverters has to have a raw type, forcing its use to have an unchecked call
+  @SuppressWarnings("unchecked")
+  public static Condition getSqlCondition(FqlCondition<?> fqlCondition, EntityType entityType) {
+    final Field<Object> field = fqlCondition instanceof FieldCondition<?> fieldCondition
+      ? field(fieldCondition, entityType)
+      : null;
```
mweaver-ebsco, line 80: This is one of the changes that caused this PR to be so big: getting the field object is somewhat expensive and almost every method was doing it at least once, so I changed it so that we get the field object right off the bat, then pass it through to all of the methods that need it.

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -68,145 +72,168 @@ public Condition getSqlCondition(String fqlCriteria, EntityType entityType) {
...
+    return sqlConverters.getOrDefault(fqlCondition.getClass(), (condition, et, f) -> falseCondition()).handle(fqlCondition, entityType, field);
   }
 
-  private Condition handleEquals(EqualsCondition equalsCondition, EntityType entityType) {
+  private static Condition handleEquals(EqualsCondition equalsCondition, EntityType entityType, Field<Object> field) {
```
mweaver-ebsco, line 84: This is the other tiny change that makes this PR kind of obnoxious: I made almost everything `static`. It's much more efficient and generally makes things a little easier to understand (there's no state outside the method to worry about coming from external dependencies)

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -68,145 +72,168 @@ public Condition getSqlCondition(String fqlCriteria, EntityType entityType) {
...
+  }
+
+  /**
+   * Create a case-insensitive comparison of the given fieldCondition with the given entityType.
+   * <p /><b>Note: This method always converts the comparison value to lower-case, while the column will be converted
+   * to lower-case only if it does not have a filterValueGetter (currently, we assume the filterValueGetter
+   * will convert the column to lower-case. If that assumption changes, this method will need to be updated)</b>
+   *
+   * @param toCaseInsensitiveCondition The function to build a Condition where the column is converted to lower-case
+   * @param toCaseSensitiveCondition   The function to build a Condition where the column is *not* converted to lower-case
```
mweaver-ebsco, line 225: I only bothered documenting these 2 params, since the others seemed pretty obvious (they are the same as in pretty much every other method in the class)

mweaver-ebsco: I supposed `value` isn't used everywhere, so it's not as obvious. I'll add something for that one, too

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -218,4 +245,19 @@ private static Field<Object> field(FieldCondition<?> condition, EntityType entit
...
+   * methods than we can on an object
+   */
+  private static <T extends FqlCondition<?>> Map.Entry<Class<T>, ? extends ConditionHandler<T>> buildMapping(Class<T> conditionClass, ConditionHandler<T> f) {
+    return Map.entry(conditionClass, f);
+  }
+
+  @FunctionalInterface
+  private interface ConditionHandler<T extends FqlCondition<?>> {
+    Condition handle(T fqlCondition, EntityType entityType, Field<Object> field);
+  }
```
mweaver-ebsco, line 262: Do you guys think this interface deserves a comment? It's so simple that I'm not sure if that would just be noise or not...

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
ncovercash: > so big
I guess I haven't given you enough of my PRs yet...

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
ncovercash: I think it's good without!

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
mweaver-ebsco: I guess "noisy" is probably a better description :D

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -34,17 +33,22 @@
 @Service
```
ncovercash, line 33: Now that everything in here is static, should this be a `Service`? IMO it could be a util class now...

ncovercash (approved):

bvsharp (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
mweaver-ebsco: I had the same thought, but there's still the one non-`static` method that depends on the `FqlService` :(

PR #88, opened 2023-12-19
MODFQMMGR-97:Query failures with long failure reasons result in a SQL…

kjain110:
https://issues.folio.org/browse/MODFQMMGR-97


ncovercash (approved):

ncovercash (changes requested):
this should go in 1.1.0

ncovercash (approved):

PR #89, opened 2023-12-29
MODFQMMGR-98 Add filterValueGetters to the Items entity type

mweaver-ebsco:


src/main/resources/db/changelog/changes/v1.1.0/update-drv-item-details-definition.xml
```diff
@@ -0,0 +1,330 @@
...
+              },
+              "valueGetter": "jsonb_path_query_first(instance_details.jsonb, '$.\"contributors\"[*]?(@.\"primary\" == true).\"name\"'::jsonpath) #>> '{}'::text[]",
+              "visibleByDefault": false
+            },
+            {
+              "name": "instance_title",
+              "dataType": {
+                "dataType": "stringType"
+              },
+              "filterValueGetter": "\"left\"(lower(${tenant_id}_mod_inventory_storage.f_unaccent(instance_details.jsonb ->> 'title'::text)), 600)",
```
mweaver-ebsco, line 52: The big change here (since it's hard to spot in files like this where we just have 1 big new version of the file) is the addition of these `filterValueGetter` properties on some columns. The only other change to the entity type definition is the removal of the `subEntityTypeIds` property (which isn't used anymore)

mweaver-ebsco: For context, in case someone stumbles across the PR later: the filterValueGetter is the thing we use in the WHERE clause when we convert the FQL query to SQL. This column here illustrates it nicely: there's an index that pretty much exactly matches the `filterValueGetter` here. If we use the `valueGetter` in the WHERE clause, it doesn't have any of the extra processing that is used in the index, so the index isn't used and we end up doing a seq scan in the query plan, which can be extremely slow. But we can't just change the `valueGetter` to resemble the index because then users would see values that are truncated, lowercase, and unaccented. So instead, we have 2 different value getters for rendering (`valueGetter`) vs the behind-the-scenes stuff like filtering and sorting (`filterValueGetter`)

kjain110 (approved):

ncovercash (approved):

PR #90, opened 2024-01-03
MODFQMMGR-93: fixing POL entity type

kjain110:
https://issues.folio.org/browse/MODFQMMGR-93

Screen Shot 2024-01-03 at 12 23 03 PM
Screen Shot 2024-01-03 at 12 23 13 PM


src/main/resources/db/changelog/changes/v1.0.4/changelog-v1.0.4.xml
mweaver-ebsco, line 1: Nit: These should probably be moved to the v1.1.0 changelog, since this is `master` (and if we release a v1.0.4, we probably don't want the POL entity type in it)

src/test/java/org/folio/fqm/ModFqmManagerIT.java
```diff
@@ -125,6 +125,6 @@ private void smokeTest() {
       .then()
       .statusCode(200)
       // TODO: below value should be updated to 4 once POL entity type is fixed
```
mweaver-ebsco, line 127: Nit: if this is TODOne, so we don't need the comment anymore

src/main/resources/db/changelog/changes/v1.0.4/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -39,44 +39,15 @@
...
-              "valueGetter": "COALESCE(json_array_elements.value ->> 'fundId', NULL::text)",
-              "visibleByDefault": false
+                "name": "fund_distribution",
+                "dataType": {
+                  "dataType": "arrayType",
+                  "itemDataType": {
+                   "dataType": "objectType"
+                  }
+                },
+                "valueGetter": "src_purchase_order_line.jsonb ->> 'fund_distribution'",
```
mweaver-ebsco, line 49: Out of scope for this PR, but we should start thinking about what querying from this column might look like (e.g., if someone wants to query based on fund distribution code).

ncovercash (approved):

src/main/resources/db/changelog/changes/v1.0.4/changelog-v1.0.4.xml
kjain110: @mweaver-ebsco  the POL stuff has been moved to v1.1.0

mweaver-ebsco (approved):

src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
mweaver-ebsco, line 1: Something that just occurred to me: we need to make sure we have a dependency in the module descriptor for this!

kjain110: @mweaver-ebsco  Added

descriptors/ModuleDescriptor-template.json
```diff
@@ -232,13 +232,29 @@
...
       "version": "7.1"
     },
     {
-      "id": "loan-policy-storage",
-      "version": "2.3"
+      "id": "organizations",
+      "version": "3.3"
+    },
+    {
+      "id": "po-lines",
```
bvsharp, line 248: I think this should be "orders-storage.po-lines". We may also want to use the version from the latest release (12.2), rather than the one on master.

PR #93, opened 2024-01-05
MODFQMMGR-102 Rearrange the columns in the query_results PK

mweaver-ebsco:
This improves the performance for nearly all queries that we do with that table by changing it to (query_id, result_id), since in pretty much every query, we provide a query ID and no result IDs

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
@@ -14,4 +14,12 @@
   <include file="update-src-organizations-definition.xml" relativeToChangelogFile="true"/>
   <include file="update-drv-purchase-order-line-details-definition.xml" relativeToChangelogFile="true"/>
 
+  <changeSet id="MODFQMMGR-107" author="mweaver@ebsco.com">
+    <comment>Swap the columns in the query_results primary key, to improve performance</comment>
+    <sql>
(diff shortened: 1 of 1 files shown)


ncovercash (approved):
🏎️

bvsharp (approved):

src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
```diff
@@ -14,4 +14,12 @@
   <include file="update-src-organizations-definition.xml" relativeToChangelogFile="true"/>
   <include file="update-drv-purchase-order-line-details-definition.xml" relativeToChangelogFile="true"/>
 
+  <changeSet id="MODFQMMGR-107" author="mweaver@ebsco.com">
+    <comment>Swap the columns in the query_results primary key, to improve performance</comment>
+    <sql>
+      ALTER TABLE query_results DROP CONSTRAINT IF EXISTS query_results_pkey;
```
kjain110, line 20: Should we add` if exists ` with Alter table 
Like ALTER TABLE IF EXISTS ??

kjain110 (approved):

src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
mweaver-ebsco: I feel like having it blow up if the table doesn't exist seems like a reasonable thing to do, since the app will be pretty broken if the table isn't there

PR #96, opened 2024-01-08
MODFQMMGR-93: fixing null values

kjain110:


src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -220,9 +220,12 @@
             {
               "name": "purchase_order_notes",
               "dataType": {
```
mweaver-ebsco, line 222: Personal opinion: I'd add a `runOnChange="true"` to the `<changeSet>` tag, to make this integrate nicely with our existing DBs (otherwise, we have to drop the DB schema or truncate `databasechangelog`)

mweaver-ebsco (approved):

PR #98, opened 2024-01-11
MODFQMMGR-113: Rework array operators to avoid substring comparisons

bvsharp:
## Purpose
[MODFQMMGR-113: Rework array operators to avoid substring comparisons](https://issues.folio.org/projects/MODFQMMGR/issues/MODFQMMGR-113)

# Notes
- This ticket supports $contains and $not_contains operators. Other operators have been discussed for use with arrays ($in, $nin), but these operators are not in the scope of this ticket
- This does not fix the POL fund distribution array column, this will be handled in another ticket

## Testing
- [x] All unit tests passed
- [x] All array fields (minus POL fund distribution) can be queried with $contains and $not_contains operators. These operators use array-contains logic and do not search for/match substrings.

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -10,6 +10,7 @@ import org.folio.querytool.domain.dto.EntityType;
 import org.folio.querytool.domain.dto.EntityTypeColumn;
 import org.jooq.Condition;
 import org.jooq.Field;
+import org.jooq.impl.DSL;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -202,17 +203,22 @@ private static Condition handleRegEx(RegexCondition regexCondition, EntityType e
 
   private static Condition handleContains(ContainsCondition containsCondition, EntityType entityType, Field<Object> field) {
     if (containsCondition.value() instanceof String value) {
-      return caseInsensitiveComparison(containsCondition, entityType, field, value, Field::containsIgnoreCase, Field::contains);
```
bvsharp, line 205: I was originally trying to get the new logic working with the existing caseInsensitiveComparison method but it ended up becoming pretty messy due to the different data types. In the end I figured it would be easier to just cast the value to lowercase within the handleContains/handleNotContains methods themselves.

ncovercash (approved):

PR #99, opened 2024-01-11
MODFQMMGR-110 Add filterValueGetters to the Loans entity type

mweaver-ebsco:


src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
```diff
@@ -0,0 +1,352 @@
...
+          "private": false,
+          "fromClause": "src_circulation_loan LEFT JOIN src_circulation_loan_policy loan_policy_ref_data ON loan_policy_ref_data.id = ((src_circulation_loan.jsonb ->> 'loanPolicyId'::text)::uuid) LEFT JOIN src_inventory_service_point cospi ON cospi.id = \"left\"(lower(src_circulation_loan.jsonb ->> 'checkoutServicePointId'::text), 600)::uuid LEFT JOIN src_inventory_service_point cispi ON cispi.id = \"left\"(lower(src_circulation_loan.jsonb ->> 'checkinServicePointId'::text), 600)::uuid JOIN src_inventory_item item_details ON item_details.id = \"left\"(lower(f_unaccent(src_circulation_loan.jsonb ->> 'itemId'::text)), 600)::uuid LEFT JOIN src_inventory_material_type material_type_ref_data ON material_type_ref_data.id = item_details.materialtypeid JOIN src_users_users user_details ON user_details.id = \"left\"(lower(f_unaccent(src_circulation_loan.jsonb ->> 'userId'::text)), 600)::uuid LEFT JOIN src_users_groups patron_id_ref_data ON patron_id_ref_data.id = user_details.patrongroup JOIN src_inventory_holdings_record hrim ON item_details.holdingsrecordid = hrim.id JOIN src_inventory_instance instance_details ON hrim.instanceid = instance_details.id",
```
mweaver-ebsco, line 31: I modified the join condition on `patron_id_ref_data`, to make it use `user_details.patrongroup` instead of `user_details.jsonb ->> 'patronGroup'`. Both versions of the patron group have indexes, but the `jsonb` one requires a function. Since it's a `join`, using the `jsonb` one could mean calling that function for every row in `user_details` in order to compare it to `patron_id_ref_data.id`, which could could expensive. Using the plain column is pretty much free

src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
mweaver-ebsco, line 1: Most of the changes in this file happen in separate commits, so you can actually see small diffs for most changes if you switch the PR view from "all commits" to individual commits. AFAIK, the only change that will get lost in the noise with that approach is the `instance_title` filterValueGetter, since that was added in the same commit as the one that created the file

mweaver-ebsco: The only changes that should be here are the change to the `FROM` clause and the addition of `filterValueGetters`; everything else in this entity type def should be unchanged

ncovercash (approved):

bvsharp (approved):

src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
```diff
@@ -0,0 +1,352 @@
...
+            {
+              "name": "loan_checkout_servicepoint_id",
+              "dataType": {
+                "dataType": "rangedUUIDType"
+              },
+              "valueGetter": "cospi.id",
+              "visibleByDefault": false
+            },
+            {
+              "name":,
```
mweaver-ebsco, line 167: Something bad happened here... This isn't valid JSON

src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
```diff
@@ -0,0 +1,352 @@
...
+              "dataType": {
+                "dataType": "stringType"
+              },
+              "idColumnName": "loan_checkout_servicepoint_id",
+              "filterValueGetter": "lower(${tenant_id}_mod_inventory_storage.f_unaccent(cospi.jsonb ->> 'name'::text))",
+              "valueGetter": "cospi.jsonb ->> 'name'",
+              "visibleByDefault": false
+            },
+            {
+              "name": "g",
```
mweaver-ebsco, line 181: This name got mangled somehow

src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
mweaver-ebsco: I think this one should be `loan_checkout_servicepoint_name`

src/main/resources/db/changelog/changes/v1.1.0/update-drv-loan-details-definition.xml
mweaver-ebsco: I think this one should be `loan_due_date`

kjain110 (approved):

PR #102, opened 2024-01-18
MODFQMMGR-99:Support for custom fields

kjain110:
**The Jenkins will pass after we merge the code for folio-query-tool-metadata**

### DIFF
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -35,6 +35,10 @@ import static org.jooq.impl.DSL.*;
 public class EntityTypeRepository {
   public static final String ID_FIELD_NAME = "id";
   public static final String TABLE_NAME = "entity_type_definition";
+  public static final  String requiredFieldName = "jsonb ->> 'name'";
+  public static final String refId =  "jsonb ->> 'refId'";
+  public static final String whereClauseName = "jsonb ->> 'type'";
+  public static final String customFieldType = "SINGLE_CHECKBOX";
   private final DSLContext jooqContext;
   private final ObjectMapper objectMapper;
 
@@ -53,38 +57,20 @@ public class EntityTypeRepository {
   public List<EntityTypeColumn> fetchNamesForSingleCheckbox(UUID entityTypeId) {
     log.info("Getting derived table name for entity type ID: {}", entityTypeId);
     String sourceViewName = getDerivedTableName(entityTypeId).get();
-    List<EntityTypeColumn> entityColumnsList = new ArrayList<>();
 
-    String requiredFieldName = "jsonb ->> 'name'";
-    String refId =  "jsonb ->> 'refId'";
-    String whereClauseName = "jsonb ->> 'type'";
-    String customFieldType = "SINGLE_CHECKBOX";
-
-    ValueWithLabel valueWithLabelTrue = new ValueWithLabel().label("True").value("true");
-    ValueWithLabel valueWithLabelFalse = new ValueWithLabel().label("False").value("false");
-
-    Result<Record2<Object, Object>> result = jooqContext
+    return jooqContext
       .select(field(requiredFieldName), field(refId))
       .from(sourceViewName)
       .where(field(whereClauseName).eq(customFieldType))
-      .fetch();
-
-    result.stream()
+      .fetch()
+      .stream()
       .map(record -> {
         Object value = record.get(0);
-        Object refSelect = record.get(1);
-        EntityTypeColumn entityTypeColumn = new EntityTypeColumn();
+        Object refId1 = record.get(1);
         assert value != null;
-        entityTypeColumn.name(value.toString());
-        entityTypeColumn.dataType(new BooleanType());
-        entityTypeColumn.values(List.of(valueWithLabelTrue, valueWithLabelFalse));
-        entityTypeColumn.visibleByDefault(false);
-        entityTypeColumn.valueGetter("src_users_users.jsonb -> 'customFields' ->> '"+refSelect+"'");
-        return entityTypeColumn;
+        return createEntityTypeColumn(value.toString(), refId1.toString());
       })
-      .forEach(entityColumnsList::add);
-
-    return entityColumnsList;
+      .collect(Collectors.toList());
   }
 
 
@@ -124,6 +110,20 @@ public class EntityTypeRepository {
       .map(
         row -> new RawEntityTypeSummary(row.get(idField), row.get(nameField))
       );
+  }
+
+  private EntityTypeColumn createEntityTypeColumn(String value, String refId) {
+    ValueWithLabel trueValue = new ValueWithLabel().label("True").value("true");
+    ValueWithLabel falseValue = new ValueWithLabel().label("False").value("false");
+
+    EntityTypeColumn entityTypeColumn = new EntityTypeColumn();
+    entityTypeColumn.name(value);
+    entityTypeColumn.dataType(new BooleanType());
+    entityTypeColumn.values(List.of(trueValue, falseValue));
+    entityTypeColumn.visibleByDefault(false);
+    entityTypeColumn.valueGetter("src_users_users.jsonb -> 'customFields' ->> '" + refId + "'");
+
+    return entityTypeColumn;
   }
 
   @SneakyThrows
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -4,7 +4,6 @@ import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.extern.log4j.Log4j2;
 
-import org.folio.fqm.exception.ColumnNotFoundException;
 import org.folio.fqm.exception.EntityTypeNotFoundException;
 import org.folio.querytool.domain.dto.EntityType;
 import org.folio.querytool.domain.dto.EntityTypeColumn;
@@ -13,8 +12,6 @@ import org.folio.querytool.domain.dto.BooleanType;
 import org.jooq.DSLContext;
 import org.jooq.Condition;
 import org.jooq.Field;
-import org.jooq.Result;
-import org.jooq.Record2;
 import org.springframework.stereotype.Repository;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -23,7 +20,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.List;
-import java.util.ArrayList;
 import java.util.stream.Collectors;
 
 import static org.apache.commons.collections4.CollectionUtils.isEmpty;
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -4,26 +4,18 @@
 import lombok.SneakyThrows;
 import lombok.extern.log4j.Log4j2;
 
-import org.jooq.DSLContext;
-import org.jooq.Condition;
-import org.jooq.Field;
+import org.folio.querytool.domain.dto.*;
```
ncovercash, line 7: Please avoid wildcard imports

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -14,12 +14,7 @@
 
 import static org.folio.fqm.repository.EntityTypeRepository.ID_FIELD_NAME;
 
-import java.util.Comparator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
```
ncovercash, line 22: Please avoid wildcard imports

src/main/resources/application.yml
```diff
@@ -11,9 +11,9 @@ spring:
       password: ${DB_PASSWORD:postgres}
       url: jdbc:postgresql://${DB_HOST:localhost}:${DB_PORT:5432}/${DB_DATABASE:db}
     reader:
-      username: ${DB_USERNAME:postgres}
-      password: ${DB_PASSWORD:postgres}
-      url: jdbc:postgresql://${DB_HOST_READER:localhost}:${DB_PORT_READER:5432}/${DB_DATABASE:db}
+        username: ${DB_USERNAME:postgres}
+        password: ${DB_PASSWORD:postgres}
```
ncovercash, line 15: Did you mean to commit this?

ncovercash (changes requested):

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
       .fetchOptional(derivedTableNameField);
   }
 
+  public List<EntityTypeColumn> fetchNamesForSingleCheckbox(UUID entityTypeId) {
+    log.info("Getting derived table name for entity type ID: {}", entityTypeId);
+    String sourceViewName = getDerivedTableName(entityTypeId).get();
+    List<EntityTypeColumn> entityColumnsList = new ArrayList<>();
+
+    String requiredFieldName = "jsonb" + "->>'name'";
+    String abc =  "jsonb" + "->>'refId'";
```
mweaver-ebsco, line 58: This variable should probably have a new name

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+    String whereClauseName = "jsonb" + "->>'type'";
```
mweaver-ebsco, line 59: Tiniest of nits: all of these `"jsonb" + "->>'something'"` strings might as well be made into simple string constants (i.e., `jsonb ->> 'something'`). It doesn't really matter, since the compiler should do the concatenation at compile-time, but it just looks cleaner and will probably keep sonarqube from complaining about a duplicate string

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+    String customFieldType = "SINGLE_CHECKBOX";
+    ValueWithLabel valueWithLabelTrue = new ValueWithLabel().label("True").value("true");
+    ValueWithLabel valueWithLabelFalse = new ValueWithLabel().label("False").value("false");
+
+    Result<Record2<Object, Object>> result = jooqContext
+      .select(field(requiredFieldName), field(abc))
+      .from(sourceViewName)
+      .where(field(whereClauseName).eq(customFieldType))
```
mweaver-ebsco, line 67: Rather than rebuilding these fields here every time this method is called, maybe store the `Field` objects as `static` fields on the class

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+
+    String requiredFieldName = "jsonb" + "->>'name'";
+    String abc =  "jsonb" + "->>'refId'";
+    String whereClauseName = "jsonb" + "->>'type'";
+    String customFieldType = "SINGLE_CHECKBOX";
+    ValueWithLabel valueWithLabelTrue = new ValueWithLabel().label("True").value("true");
+    ValueWithLabel valueWithLabelFalse = new ValueWithLabel().label("False").value("false");
+
+    Result<Record2<Object, Object>> result = jooqContext
+      .select(field(requiredFieldName), field(abc))
```
mweaver-ebsco, line 65: I don't know for sure, but I suspect we could tighten up the types here by setting types on the `Field` objects in the `select()` call (e.g., `.select(field(requiredFieldName, String.class), field(ABC, String.class))`)

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
       .fetchOptional(derivedTableNameField);
   }
 
+  public List<EntityTypeColumn> fetchNamesForSingleCheckbox(UUID entityTypeId) {
```
mweaver-ebsco, line 52: The name of this method should probably reflect that it's for custom fields

mweaver-ebsco, line 52: Rather than fetch only the checkbox custom fields (implying that we'll need to do a separate method for radios vs text inputs vs etc, where each is `select ..., ... from ... where type = ______`), I'd do 1 method for all custom fields (so that we only do 1 query of `select ..., ..., type from ...`), then iterate over the results of that query, dispatching to different methods, depending on the type of the custom field, skipping+logging types that we don't support yet

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+    Result<Record2<Object, Object>> result = jooqContext
+      .select(field(requiredFieldName), field(abc))
+      .from(sourceViewName)
+      .where(field(whereClauseName).eq(customFieldType))
+      .fetch();
+    result.stream()
+      .map(record -> {
+        Object value = record.get(0);
+        EntityTypeColumn entityTypeColumn = new EntityTypeColumn();
+        entityTypeColumn.name((value instanceof List) ?
```
mweaver-ebsco, line 73: Is this ever a `List`? If I'm understanding the code, it's this just the name of the custom field?

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+          ((List<?>) value).stream().map(item -> item != null ? item.toString() : null).collect(Collectors.joining(", "))
```
mweaver-ebsco, line 74: If `value` _can_ be a `List`, then I'd use pattern matching to remove the cast here

```suggestion
        entityTypeColumn.name((value instanceof List valueList) ?
          valueList.stream().map(item -> item != null ? item.toString() : null).collect(Collectors.joining(", "))
```

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -47,6 +49,40 @@ public Optional<String> getDerivedTableName(UUID entityTypeId) {
...
+          : (value != null ? value.toString() : null));
+        entityTypeColumn.dataType(new BooleanType());
+        entityTypeColumn.values(List.of(valueWithLabelTrue, valueWithLabelFalse));
+        entityTypeColumn.visibleByDefault(false);
+        entityTypeColumn.valueGetter("user_custom_fields.jsonb -> 'customFields' ->> 'refId'");
+        return entityTypeColumn;
+      })
+      .forEach(entityColumnsList::add);
+    return entityColumnsList;
```
mweaver-ebsco, line 83: Rather than create a list, fill it, then return it, I'd do it all in 2 fluid chain of calls

So return the result of
```suggestion
      .toList();
```
and just remove `entityColumnsList`

mweaver-ebsco: Actually, since the result is just getting added to another list, I'd just return the `Stream` directly and let the call decide what to do with it. That way, we avoid allocating and populating a list just to throw it away almost immediately; instead, the caller can just consume the stream directly without needing to allocate anything extra

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -60,10 +60,14 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
         entityType.setLabelAlias(localizationService.getEntityTypeLabel(entityType.getName()));
 
-        entityType.getColumns().forEach(column -> 
+        entityType.getColumns().forEach(column ->
           column.setLabelAlias(localizationService.getEntityTypeColumnLabel(entityType.getName(), column.getName()))
         );
-
+        String customFieldsEntityTypeId = entityType.getCustomFieldEntityTypeId();
+        if (customFieldsEntityTypeId != null) {
+          entityTypeRepository.fetchNamesForSingleCheckbox(UUID.fromString(customFieldsEntityTypeId));
```
mweaver-ebsco, line 68:
```suggestion
```

src/main/resources/db/changelog/changes/v1.1.0/harvested-views.xml
```diff
@@ -48,4 +48,15 @@
     </createView>
   </changeSet>
 
+  <changeSet id="create_view_custom_fields" author="kjain@ebsco.com">
+    <preConditions onFail="CONTINUE">
+      <tableExists tableName = "custom_fields" schemaName="${tenant_id}_mod_users"/>
+    </preConditions>
+    <createView
+      replaceIfExists="true"
+      viewName="user_custom_fields">
```
mweaver-ebsco, line 57: Nit: we should put some kind of prefix on this view name, to indicate that it's not one of the core FQM DB objects. `src_` could make sense, since this view is definitely source data. But it's not like other source data, since it defines columns instead of actual data, so maybe something else...

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
kjain110: @mweaver-ebsco  yes you are right it just a name not a list. Yesterday there was some thing With the name so it wasn't working properly. No Its all good. I guess we don't need to write it as list and all that code.
We can just have `entityTypeColumn.name(value.toString());`

PR #104, opened 2024-01-22
MODFQMMGR-125: Add _deleted field for content items that have been deleted

bvsharp:
## Purpose
[MODFQMMGR-125](https://issues.folio.org/browse/MODFQMMGR-125)

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,15 @@ Example:
 }
 ```
 
+## FQL reserved field names
+Field names may be prefixed with special characters to indicate that the field is reserved for a specific purpose.
+
+### Underscore (_)
+Fields prefixed with an underscore are reserved for use by FQM.
+
+#### _deleted
+The field name "_deleted" can be used to indicate that the record corresponding to a particular content ID has been deleted.
+
 ## Additional information
 
 ### Issue tracker
--- a/src/main/java/org/folio/fqm/service/ResultSetService.java
+++ b/src/main/java/org/folio/fqm/service/ResultSetService.java
@@ -33,7 +33,7 @@ public class ResultSetService {
       .collect(Collectors.toMap(content -> (UUID) content.get(ID_FIELD_NAME), Function.identity()));
 
     return contentIds.stream()
-      .map(contentsMap::get)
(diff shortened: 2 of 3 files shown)


src/main/java/org/folio/fqm/service/ResultSetService.java
```diff
@@ -33,7 +33,7 @@ private static List<Map<String, Object>> getSortedContents(List<UUID> contentIds
       .collect(Collectors.toMap(content -> (UUID) content.get(ID_FIELD_NAME), Function.identity()));
 
     return contentIds.stream()
-      .map(contentsMap::get)
+      .map(id -> contentsMap.getOrDefault(id, Map.of("id", id, "_deleted", true)))
```
mweaver-ebsco, line 36: Small thing, but I'd move this `Map.of()` call outside the lambda. Otherwise, we end up allocating/populating an identical `Map` for every single row

README.md
```diff
@@ -377,6 +377,15 @@ Example:
 }
 ```
 
+## FQL reserved field names
+Field names may be prepended with special characters to indicate that the field is reserved for a specific purpose.
```
ncovercash, line 381: Nit: prefixed may be a better word

ncovercash (approved):

README.md
```diff
@@ -377,6 +377,15 @@ Example:
...
+
+### Underscore (_)
+Fields prepended with an underscore are reserved for use by FQM consumers.
```
mweaver-ebsco, line 384: I misunderstood the implementation in chat (I thought mod-lists would look for missing records in its results from FQM and inject the missing ones; the approach here works, too, though). Since this actually _is_ being set by FQM, then we should reserve the `_` prefix for use by FQM instead of FQM consumers

src/main/java/org/folio/fqm/service/ResultSetService.java
bvsharp: We'd have to do it here because the id would make each map unique right?

mweaver-ebsco: Oh yeah, good point 😆

PR #105, opened 2024-01-22
MODFQMMGR-116 Add support for value functions on fields

mweaver-ebsco:
This makes it so we can configure a "valueFunction" property on a column
within an entity type, allowing us to customize how query conditions on
a column work. This way, if a filter value getter on a field changes its
behavior (e.g., removing accents), we can make the condition's value
exhibit that same behavior, to make it more intuitive to work with.

An example of this is Loan.instance_title: the filter value getter on
that field converts it to lower-case, truncates it to 600 characters,
and removes all accents. This makes it so a user could copy/paste a
title with accents from query results directly into an `instance_title
== <the title>` query and not get the expected result. With this commit,
we can do the same `lower(truncate(unaccent()))` processing on both the
field (using the filter value getter) and on the condition's static
value (using the value function) and get more consistent results.

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -23,9 +23,11 @@ import java.util.Map;
 import java.util.function.BiFunction;
 
 import static org.jooq.impl.DSL.and;
+import static org.jooq.impl.DSL.cast;
 import static org.jooq.impl.DSL.condition;
 import static org.jooq.impl.DSL.falseCondition;
 import static org.jooq.impl.DSL.or;
+import static org.jooq.impl.DSL.param;
 import static org.jooq.impl.DSL.val;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -253,6 +253,17 @@ private static Field<Object> field(FieldCondition<?> condition, EntityType entit
       .orElseThrow(() -> new ColumnNotFoundException(entityType.getName(), columnName));
   }
 
+  // Suppress the unchecked cast warning on the Class<T> cast below. We need the correct type there in order to get
+  // a reasonable level of type-safety with the return type on this method
+  @SuppressWarnings("unchecked")
```
mweaver-ebsco, line 258: I fought with this for a while and eventually just gave up on trying to do it right and just did a cast to `Class<T>` in the method.

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceTest.java
```diff
@@ -288,6 +296,40 @@ static List<Arguments> jooqConditionsSource() {
...
+      ),
+
+      Arguments.of(
+        "in operator on a field with a valueFunction",
+        """
+          {"fieldWithAValueFunction": {"$in": ["value1", 2, true]}}""",
+        or(
+          field("fieldWithAValueFunction").equalIgnoreCase(field("upper(:value)", String.class, param("value", "value1"))),
+          field("fieldWithAValueFunction").eq(field("upper(:value)", String.class, param("value", 2))),
+          field("fieldWithAValueFunction").eq(field("upper(:value)", String.class, param("value", true))))
```
mweaver-ebsco, line 321: Kind of annoying, but you have to do the named parameter binding in the expected results, otherwise the objects aren't considered equivalent

bvsharp (approved):

ncovercash (approved):

PR #107, opened 2024-01-26
MODFQMMGR-119: Add empty FQL operator

bvsharp:
## Purpose
[MODFQMMGR-119](https://issues.folio.org/browse/MODFQMMGR-119): Create new operators to support searching for null/empty values

## Testing
- [x] All unit tests passed
- [x] Empty operator works for all datatypes

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -24,6 +24,7 @@ import java.util.function.BiFunction;
 
 import static org.jooq.impl.DSL.and;
 import static org.jooq.impl.DSL.cast;
+import static org.jooq.impl.DSL.cardinality;
 import static org.jooq.impl.DSL.condition;
 import static org.jooq.impl.DSL.falseCondition;
 import static org.jooq.impl.DSL.or;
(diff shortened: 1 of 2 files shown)


ncovercash (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -221,13 +223,33 @@ private static Condition handleNotContains(NotContainsCondition notContainsCondi
     return cast(field, String[].class).notContains(valueField(new String[]{notContainsCondition.value().toString()}, notContainsCondition, entityType));
   }
 
+  private static Condition handleEmpty(EmptyCondition emptyCondition, EntityType entityType, Field<Object> field) {
+    String fieldType = getColumn(emptyCondition, entityType)
+      .getDataType()
+      .getDataType();
+    boolean isEmpty = Boolean.TRUE.equals(emptyCondition.value());
+    var nullCondition = isEmpty ? field.isNull() : field.isNotNull();
+    switch (fieldType) {
```
ncovercash, line 232:
```suggestion
    return switch (fieldType) {
      case "stringType" -> isEmpty ? nullCondition.or(field.eq("")) : nullCondition.and(field.ne("")),
      case "arrayType" -> {
        var cardinality = cardinality(DSL.cast(field, String[].class));
        yield isEmpty ? nullCondition.or(cardinality.eq(0)) :  nullCondition.and(cardinality.ne(0));
      },
      default -> nullCondition,
    }
```

PR #110, opened 2024-01-31
Modfqmmgr 101

vgema:
## Purpose

## Approach

#### TODOS and Open Questions

## Learning


src/main/resources/db/changelog/changes/v1.1.0/sql/insert-item-status-definition.sql
mweaver-ebsco, line 1: Why create a whole table to hold the enum values instead of just putting them directly into the entity type definition? To do that, we just need to add a `values` property to the column definition, which is an array containing `{"label": string, "value": string}` objects

src/test/java/org/folio/fqm/utils/ItemStatusEnumTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+      if (responseCode == HttpURLConnection.HTTP_OK) {
+        BufferedReader in = new BufferedReader(new InputStreamReader(connection.getInputStream()));
+        String inputLine;
+        StringBuilder content = new StringBuilder();
+
+        while ((inputLine = in.readLine()) != null) {
+          content.append(inputLine);
+        }
+        in.close();
```
mweaver-ebsco, line 35: Tiny thing, but `BufferedReader` implements `AutoCloseable`, so it'd be a little better to do this in a try-with-resources block. It avoids the call to `close()` while also guaranteeing that `close()` definitely gets called, even if something throws an exception:

```suggestion
        try (BufferedReader in = new BufferedReader(new InputStreamReader(connection.getInputStream()))) {
          String inputLine;
          StringBuilder content = new StringBuilder();

          while ((inputLine = in.readLine()) != null) {
            content.append(inputLine);
          }
        }
```

src/test/java/org/folio/fqm/utils/ItemStatusEnumTest.java
```diff
@@ -0,0 +1,65 @@
...
+      int responseCode = connection.getResponseCode();
+
+      if (responseCode == HttpURLConnection.HTTP_OK) {
+        BufferedReader in = new BufferedReader(new InputStreamReader(connection.getInputStream()));
+        String inputLine;
+        StringBuilder content = new StringBuilder();
+
+        while ((inputLine = in.readLine()) != null) {
+          content.append(inputLine);
+        }
```
mweaver-ebsco, line 34: Whenever I see something like this, I wonder what it'd look like as a `for`-loop...

```java
        for (String inputLine = in.readLine(); inputLine != null; inputLine = in.readLine()) {
          content.append(inputLine);
        }
```

...that looks kind of gross. `while`-loop it is! 🙂

mweaver-ebsco: Even better: use a stream! `BufferedReader.lines()` returns a `Stream<String>`, which can be collected directly into  a new string.

```suggestion
        String content = in.lines().collect(Collectors.joining());
```

src/test/java/org/folio/fqm/utils/ItemStatusEnumTest.java
mweaver-ebsco: Note: This puts `content` into the `try`-block's scope, so that would need to be fixed. This makes me really wish java supported something like
```
String content = () -> { ... do stuff; return something; };
```

...oh how I wish we could use Kotlin...

PR #111, opened 2024-02-01
MODFQMMGR-121:POL adjust fields in the POL entity type

kjain110:
## Purpose
Add filterValueGetter for POL entity type


ncovercash (approved):

src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -262,26 +262,27 @@
                 "dataType": "stringType"
               },
               "valueGetter": "purchase_order.jsonb ->> 'orderType'",
+              "filterValueGetter": "\"left\"(lower(${tenant_id}_mod_orders_storage.f_unaccent(purchase_order.jsonb ->> 'orderType'::text)), 600)",
```
bvsharp, line 265: We should add valueFunctions everywhere where we have filterValueGetters. For example, right here, we can have:

"valueFunction": "\"left\"(lower(${tenant_id}_mod_orders_storage.f_unaccent(:value)), 600)",

src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -246,11 +246,11 @@
               "visibleByDefault": true
             },
             {
-              "name": "purchase_order_notes",
+              "name": "PO notes",
```
bvsharp, line 249: These should keep the underscore format, so like "po_notes". And then the backend name will have to be changed in the translations files. As it is right now, I don't think the translations will work properly (because there's no "PO notes" key in the translations file)

PR #114, opened 2024-02-01
MODFQMMGR-116 Add integration test for valueFunctions

mweaver-ebsco:


src/test/java/org/folio/fqm/IntegrationTestBase.java
```diff
@@ -0,0 +1,123 @@
...
+    // running before any other integration tests run
+
+    // Test: migrate from 0.0.0, migration should be idempotent
+    body = """
+           {"module_from": "0.0.0",
+           """ + body.substring(1);
+    postTenant(body);
+
+    // Test: smoke test the /entity-types endpoint to verify DB interactions are working
+    smokeTest();
```
mweaver-ebsco, line 85: All of this was in tests before. I moved it here so that it only runs exactly once (vs keeping it in a `@Test` in the base class, which would cause it to run for every class that inherits the base class)

src/test/java/org/folio/fqm/IntegrationTestBase.java
```diff
@@ -0,0 +1,123 @@
...
+    RestAssured.requestSpecification = null;
+  }
+
+  protected static DataSource getDataSource() {
+    final HikariConfig config = new HikariConfig();
+    config.setJdbcUrl(postgres.getJdbcUrl());
+    config.setUsername(postgres.getUsername());
+    config.setPassword(postgres.getPassword());
+    return new HikariDataSource(config);
+  }
```
mweaver-ebsco, line 99: This method is new, so that inheriting classes can get direct access to the DB

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceIT.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.UUID;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.is;
+
+@Testcontainers
+class FqlToSqlConverterServiceIT extends IntegrationTestBase {
+
+  @Test
+  void valueFunctionTest() throws JsonProcessingException {
```
mweaver-ebsco, line 25: I originally had this split between a few methods, but so much of it is specific to this test that I figured it's okay to keep as one for now. We can always split it up later when we add more tests, since then it will be more clear how we should do that

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceIT.java
```diff
@@ -0,0 +1,95 @@
...
+          .visibleByDefault(true)
+          .valueGetter("some_column")
+          .filterValueGetter("lower(\"left\"(some_column, 10))")
+          .valueFunction("lower(\"left\"(:value, 10))")
+      ))
+      .fromClause("""
+        (values
+          ('2af997b6-2655-459e-bdca-decbf54795ae'::uuid, 'AbCdEfGhIjKlMnOpQrStUvWxYz', 456),
+          ('e0e4233e-fea0-4834-96ac-78739a1856d3'::uuid, 'blah blah blah', 789)
+        ) as t (id, some_column, unused_column)""");
```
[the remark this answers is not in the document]
bvsharp: neat! I never thought about how you could just bake the data into the from clause

bvsharp (approved):

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceIT.java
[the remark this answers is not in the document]
mweaver-ebsco: It works really well until we implement dynamically building the `FROM` clause. Then, we'll either need to update this test to do something different or just build in a way to let us keep doing it like this for tests

ncovercash (approved):

PR #115, opened 2024-02-01
MODFQMMGR-112-POL - adjust fields in the POL entity type - part 1

kjain110:


ncovercash (approved):

translations/mod-fqm-manager/en.json
```diff
@@ -200,7 +200,7 @@
 
   "entityType.drv_purchase_order_line_details": "Purchase order Lines",
   "entityType.drv_purchase_order_line_details.acqunit_ids": "Acquisition unit ID list",
-  "entityType.drv_purchase_order_line_details.acqunit_names": "Acquisition unit list",
+  "entityType.drv_purchase_order_line_details.acquisition_unit": "Acquisition unit",
```
bvsharp, line 203: I know the ticket says to call this "Acquisition unit", but with the array columns, we had agreed to have "list" on the end to indicate that it's a list and not a single value. We may want to run this one by Kathleen.

bvsharp (approved):

PR #120, opened 2024-02-05
 kriti/MODFQMMGR-99: adding user custom fields

kjain110:
Screen Shot 2024-02-06 at 1 40 43 AM

### DIFF
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -4,6 +4,10 @@ import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.extern.log4j.Log4j2;
 
+import org.folio.fqm.exception.EntityTypeNotFoundException;
+import org.folio.querytool.domain.dto.BooleanType;
+import org.folio.querytool.domain.dto.EntityTypeColumn;
+import org.folio.querytool.domain.dto.ValueWithLabel;
 import org.jooq.DSLContext;
 import org.jooq.Condition;
 import org.jooq.Field;
@@ -31,7 +35,10 @@ import static org.jooq.impl.DSL.trueCondition;
 public class EntityTypeRepository {
   public static final String ID_FIELD_NAME = "id";
   public static final String TABLE_NAME = "entity_type_definition";
-
+  public static final String REQUIRED_FIELD_NAME = "jsonb ->> 'name'";
+  public static final String REF_ID = "jsonb ->> 'refId'";
+  public static final String TYPE_FIELD = "jsonb ->> 'type'";
+  public static final String CUSTOM_FIELD_TYPE = "SINGLE_CHECKBOX";
   private final DSLContext jooqContext;
   private final ObjectMapper objectMapper;
 
@@ -57,7 +64,14 @@ public class EntityTypeRepository {
       .from(table(TABLE_NAME))
       .where(field(ID_FIELD_NAME).eq(entityTypeId))
       .fetchOptional(definitionField)
-      .map(this::unmarshallEntityType);
+      .map(this::unmarshallEntityType)
+      .map(entityType -> {
+        String customFieldsEntityTypeId = entityType.getCustomFieldEntityTypeId();
+        if (customFieldsEntityTypeId != null) {
+          entityType.getColumns().addAll(fetchColumnNamesForCustomFields(UUID.fromString(customFieldsEntityTypeId)));
+        }
+        return entityType;
+      });
   }
 
   public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds) {
@@ -77,10 +91,48 @@ public class EntityTypeRepository {
       );
   }
 
+  private List<EntityTypeColumn> fetchColumnNamesForCustomFields(UUID entityTypeId) {
+    log.info("Getting derived table name for entity type ID: {}", entityTypeId);
+    String sourceViewName = getEntityTypeDefinition(entityTypeId)
+      .map(EntityType::getSourceView)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    return jooqContext
+      .select(field(REQUIRED_FIELD_NAME), field(REF_ID))
+      .from(sourceViewName)
+       //This where condition will be removed when handling other CustomFieldTypes.
+       //A generic method can be created with if and else statements.
+       //Currently, if implemented, "null" will be needed in the else part.
+      .where(field(TYPE_FIELD).eq(CUSTOM_FIELD_TYPE))
+      .fetch()
+      .stream()
+      .map(row -> {
+        Object value = row.get(REQUIRED_FIELD_NAME);
+        Object extractedRefId = row.get(REF_ID);
+        assert value != null : "The value is marked as non-nullable in the database";
+        return handleSingleCheckBox(value.toString(), extractedRefId.toString());
+      })
+      .toList();
+  }
+
+  private EntityTypeColumn handleSingleCheckBox(String value, String refId) {
+    ValueWithLabel trueValue = new ValueWithLabel().label("True").value("true");
+    ValueWithLabel falseValue = new ValueWithLabel().label("False").value("false");
+
+    return new EntityTypeColumn()
+      .name(value)
+      .dataType(new BooleanType())
+      .values(List.of(trueValue, falseValue))
+      .visibleByDefault(false)
+      .valueGetter("src_users_users.jsonb -> 'customFields' ->> '" + refId + "'")
+      .labelAlias(value)
+      .isCustomField(true);
+  }
+
   @SneakyThrows
   private EntityType unmarshallEntityType(String str) {
     return objectMapper.readValue(str, EntityType.class);
   }
 
-  public record RawEntityTypeSummary(UUID id, String name) {}
+  public record RawEntityTypeSummary(UUID id, String name) {
+  }
 }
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -77,10 +89,46 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
 
+  public List<EntityTypeColumn> fetchColumnNamesForCustomFields(UUID entityTypeId) {
+    log.info("Getting derived table name for entity type ID: {}", entityTypeId);
+    String sourceViewName = getEntityTypeDefinition(entityTypeId)
+      .map(EntityType::getSourceView)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    return jooqContext
+      .select(field(REQUIRED_FIELD_NAME), field(REF_ID))
+      .from(sourceViewName)
+      .where(field(WHERE_CLAUSE).eq(CUSTOM_FIELD_TYPE))
```
kjain110, line 100: This where condition will be removed when we will handle other CustomFieldTypes. We can create a generic method with if and else statement. Right now if we do so we need to "null" in the else part

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -14,10 +18,7 @@
 
 import org.folio.querytool.domain.dto.EntityType;
 
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
ncovercash, line 21: prefer non-wildcard imports

```suggestion
import java.util.List;
import java.util.Optional;
import java.util.Set;
import java.util.UUID;
```

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -52,12 +56,20 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
       .fetchOptional(definitionField)
       .map(this::unmarshallEntityType);
+
+    EntityType entityType = entityTypeOptional.orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+
+    String customFieldsEntityTypeId = entityType.getCustomFieldEntityTypeId();
+    if (customFieldsEntityTypeId != null) {
+      entityType.getColumns().addAll(fetchColumnNamesForCustomFields(UUID.fromString(customFieldsEntityTypeId)));
+    }
+    return Optional.of(entityType);
```
ncovercash, line 72: Not sure we want to handle the exception throwing here (and if we do, there's no reason to return an `Optional`).  I'd let the consuming method handle the `Optional` and leave the not found exception for the consuming method, as it was before.  An easy way to do this is with a `map`:

```suggestion
    return jooqContext
      .select(definitionField)
      .from(table(TABLE_NAME))
      .where(field(ID_FIELD_NAME).eq(entityTypeId))
      .fetchOptional(definitionField)
      .map(this::unmarshallEntityType)
      .map(entityType -> {
        String customFieldsEntityTypeId = entityType.getCustomFieldEntityTypeId();
        if (customFieldsEntityTypeId != null) {
          entityType.getColumns().addAll(fetchColumnNamesForCustomFields(UUID.fromString(customFieldsEntityTypeId)));
        }
        return entityType;
      });
```

ncovercash (changes requested):

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
mweaver-ebsco: Could you add a comment here about that? It could be a pretty subtle thing to track down later, so a big sign pointing at this line, saying "HEY!!! YOU'RE ADDING MORE CUSTOM FIELD TYPES? COOL! PAY ATTENTION TO THIS!!!!" could be invaluable later on if there's any real delay between this PR and when we do more types

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -77,10 +91,45 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
+    return jooqContext
+      .select(field(REQUIRED_FIELD_NAME), field(REF_ID))
+      .from(sourceViewName)
+      .where(field(WHERE_CLAUSE).eq(CUSTOM_FIELD_TYPE))
+      .fetch()
+      .stream()
+      .map(row -> {
+        Object value = row.get(0);
+        Object extractedRefId = row.get(1);
+        assert value != null;
```
mweaver-ebsco, line 108: I'm not sure how I feel about using `assert`... I've never actually seen it in production Java code before (mostly because it's so new), and it's new enough that I'm not sure if there's any real consensus in the Java community around it.

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -77,10 +91,45 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    return jooqContext
+      .select(field(REQUIRED_FIELD_NAME), field(REF_ID))
+      .from(sourceViewName)
+      .where(field(WHERE_CLAUSE).eq(CUSTOM_FIELD_TYPE))
+      .fetch()
+      .stream()
+      .map(row -> {
+        Object value = row.get(0);
+        Object extractedRefId = row.get(1);
```
mweaver-ebsco, line 107: Nit: I'd grab these columns by name instead of number. It's more clear and less error-prone

```suggestion
        Object value = row.get(REQUIRED_FIELD_NAME);
        Object extractedRefId = row.get(REF_ID);
```

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -31,7 +35,10 @@
 public class EntityTypeRepository {
   public static final String ID_FIELD_NAME = "id";
   public static final String TABLE_NAME = "entity_type_definition";
-
+  public static final String REQUIRED_FIELD_NAME = "jsonb ->> 'name'";
+  public static final String REF_ID = "jsonb ->> 'refId'";
+  public static final String WHERE_CLAUSE = "jsonb ->> 'type'";
```
mweaver-ebsco, line 40: This name seems misleading, since it's only a part of the where clause. I think I'd just call it "TYPE_FIELD" or something

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -77,10 +91,45 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
+  private EntityTypeColumn handleSingleCheckBox(String value, String refId) {
+    ValueWithLabel trueValue = new ValueWithLabel().label("True").value("true");
+    ValueWithLabel falseValue = new ValueWithLabel().label("False").value("false");
+
+    return new EntityTypeColumn()
+      .name(value)
+      .dataType(new BooleanType())
+      .values(List.of(trueValue, falseValue))
+      .visibleByDefault(false)
+      .valueGetter("src_users_users.jsonb -> 'customFields' ->> '" + refId + "'")
```
mweaver-ebsco, line 123: Big red flag here: custom fields aren't necessarily only in the Users module (in practice, they are, but only because no one else has bothered to use them, but they are a general concept in a shared library that could easily be used by any module), so we shouldn't be referring to `src_users_users` by name in code. That should come from an entity type definition or some other kind of config.

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -59,11 +59,13 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
         entityType.setLabelAlias(localizationService.getEntityTypeLabel(entityType.getName()));
-
-        entityType.getColumns().forEach(column -> 
-          column.setLabelAlias(localizationService.getEntityTypeColumnLabel(entityType.getName(), column.getName()))
-        );
-
+        entityType.getColumns().forEach(column -> {
+          if (Boolean.TRUE.equals(column.getIsCustomField())) {
+            column.setLabelAlias(column.getName());
+          } else {
```
mweaver-ebsco, line 65: Not a huge deal, but it might be worthwhile to add a short comment here explaining why we're not localizing custom field names (because they are user-defined at runtime, so we have no translations available)

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
kjain110: This will be handled by another ticket: https://folio-org.atlassian.net/browse/MODFQMMGR-155

src/test/resources/test-db/scripts/custom-fields-source-view.sql
```diff
@@ -0,0 +1,35 @@
...
+             "refId": "customColumn2",
+             "visible": true,
+             "entityType": "user",
+             "checkboxField": {
+               "default": false
+             }
+           }');
+
+ CREATE VIEW custom_fields_source_view AS
+ SELECT * FROM custom_fields_table;
```
mweaver-ebsco, line 35: Totally not worth changing, but worth pointing out for consideration when doing similar test views later: you can just define the view to select the `VALUES` clause directly and avoid needing to create the table

mweaver-ebsco (approved):

PR #121, opened 2024-02-05
Add unique indexes and hardcoded statuses to new materialized views

bvsharp:
## Purpose
Add unique indexes so that new materialized views can be refreshed. Also include hardcoded statuses in new views, so that they show up in the query builder even if the tenant doesn't have any records with the status.

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.1.0/sql/create-mat-view-pol-payment-status.sql
+++ b/src/main/resources/db/changelog/changes/v1.1.0/sql/create-mat-view-pol-payment-status.sql
@@ -1,5 +1,19 @@
-CREATE MATERIALIZED VIEW IF NOT EXISTS drv_pol_payment_status AS
-SELECT DISTINCT jsonb ->> 'paymentStatus' AS payment_status
-FROM ${tenant_id}_mod_orders_storage.po_line;
+DROP MATERIALIZED VIEW IF EXISTS drv_pol_payment_status;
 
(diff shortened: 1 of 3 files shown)


ncovercash (approved):

src/main/resources/db/changelog/changes/v1.1.0/sql/create-mat-view-pol-payment-status.sql
```diff
@@ -1,5 +1,19 @@
-CREATE MATERIALIZED VIEW IF NOT EXISTS drv_pol_payment_status AS
-SELECT DISTINCT jsonb ->> 'paymentStatus' AS payment_status
-FROM ${tenant_id}_mod_orders_storage.po_line;
+DROP MATERIALIZED VIEW IF EXISTS drv_pol_payment_status;
 
-REFRESH MATERIALIZED VIEW drv_pol_payment_status;
+CREATE MATERIALIZED VIEW drv_pol_payment_status AS
```
mweaver-ebsco, line 3: Does this need added to the refresh materialized view service (or wherever we have the list of materialized views hard-coded)

bvsharp: Yes, it's already there as part of my previous ticket (it's just that, because I didn't have a unique index on it originally, the refresh was failing and the API was throwing a 500 error)

mweaver-ebsco (approved):

PR #122, opened 2024-02-06
MODFQMMGR-99: adding customEntityTypeId

kjain110:


mweaver-ebsco (approved):

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -119,7 +119,7 @@ private EntityTypeColumn handleSingleCheckBox(String value, String refId) {
     ValueWithLabel falseValue = new ValueWithLabel().label("False").value("false");
 
     return new EntityTypeColumn()
-      .name(value)
+      .name("Users " +value)
```
bvsharp, line 122: very small nit, but maybe this should be "User " + value. Then it will be in line with the other columns in the user entity type.

bvsharp (approved):

PR #123, opened 2024-02-06
custom-field-users-one: correcting names

kjain110:


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -62,7 +62,7 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
         entityType.getColumns().forEach(column -> {
           //Custom field names are not localized as they are user-defined at runtime, leaving no translations available.
           if (Boolean.TRUE.equals(column.getIsCustomField())) {
-            column.setLabelAlias(column.getName());
+            column.setLabelAlias(entityType.getLabelAlias().substring(0, entityType.getLabelAlias().length() - 1)+" " +column.getName());
```
kjain110, line 65: Done to show the custom field names start with user
Given that the entity type name begins with either **src or drv**, along with others, we must acquire label aliases. These aliases, being plural, need to be converted to singular form, such as from "users" to "user." verified each entity type, noting that the names are plural while the field names are singular, so this works in all case.

ncovercash (approved):

bvsharp (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
mweaver-ebsco: > so this works in all case

This feels pretty fragile, since it depends on an undocumented assumption and only works for words that are pluralized by adding an "s" to the end. It might also run into localization issues (since we can localize entity type names and other field names, we could end up with Italian users trying to access the Utenti entity type, making it so some columns start with "Utente" and others start with "User")

Maybe instead of using the entity type label, we could use something more generic? Like "Custom Field: " (which could be localized) or a localized version of the name of the column representing the custom fields? Like, say the name is `user_custom_field`, we could localize that to something like "User custom field" and set the custom field's overall label alias to `localize(columnName) + ':' + column.getName()`

PR #130, opened 2024-02-09
MODFQMMGR-137: Change how we identify unique records for lists

bvsharp:
## Purpose
[MODFQMMGR-137](https://folio-org.atlassian.net/browse/MODFQMMGR-137): Change how we identify unique records for lists

Currently a single UUID id field is used to distinguish distinct records in FQM. This PR is to rework that logic and use an array of strings as the unique identifier instead. This will allow us to use as many fields as we like as the unique record identifier in FQM.

## Testing
- [x] All unit tests passed
- [x] Query workflow works for single or composite result ids

### DIFF
--- a/src/main/java/org/folio/fqm/repository/ResultSetRepository.java
+++ b/src/main/java/org/folio/fqm/repository/ResultSetRepository.java
@@ -10,15 +10,19 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import org.folio.fql.model.Fql;
 import org.folio.fqm.exception.ColumnNotFoundException;
 import org.folio.fqm.exception.EntityTypeNotFoundException;
 import org.folio.fqm.service.FqlToSqlConverterService;
+import org.folio.fqm.utils.IdColumnUtils;
 import org.folio.fqm.utils.SqlFieldIdentificationUtils;
+import org.folio.querytool.domain.dto.EntityDataType;
 import org.folio.querytool.domain.dto.EntityType;
 
 import org.apache.commons.collections4.CollectionUtils;
+import org.folio.querytool.domain.dto.EntityTypeColumn;
 import org.jooq.Condition;
 import org.jooq.DSLContext;
 import org.jooq.Field;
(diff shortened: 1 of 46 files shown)


src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -42,35 +42,45 @@ public class ResultSetRepository {
...
+      afterIdCondition = field(idValueGetter).greaterThan(afterIdArray);
+    } else {
+      afterIdCondition = DSL.noCondition();
+    }
+    // Make sure idColumns are included in results
+    idColumnNames.forEach(idColumn -> {
+      if (!fields.contains(idColumn)) {
+        fields.add(idColumn);
+      }
+    });
```
mweaver-ebsco, line 82: Nit: I'd either use a stream for this (so that you can do `.filter()` instead of the `if`) or do it with an old school for-each loop (`for (var idColumn : idColumnNames) {`), just to stick with a single idiom. Personally, I'd probably go old school for this one, since it's so simple and matches the style of the rest of the method

src/main/java/org/folio/fqm/service/ResultSetService.java
```diff
@@ -1,40 +1,65 @@
...
-      .collect(Collectors.toMap(content -> (UUID) content.get(ID_FIELD_NAME), Function.identity()));
-    return contentIds.stream()
+  private List<Map<String, Object>> getSortedContents(UUID entityTypeId, List<List<String>> contentIds, List<Map<String, Object>> unsortedResults) {
+    EntityType entityType = entityTypeRepository.getEntityTypeDefinition(entityTypeId)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    List<String> idColumnNames = IdColumnUtils.getIdColumnNames(entityType);
+    Map<List<String>, Map<String, Object>> contentsMap = unsortedResults.stream()
+      .collect(Collectors.toMap(content -> {
+            List<String> keys = new ArrayList<>();
+            idColumnNames.forEach(columnName -> keys.add(content.get(columnName).toString()));
```
mweaver-ebsco, line 44: Do we need to worry about `content.get(columnName)` ever returning `null` here?

src/main/java/org/folio/fqm/utils/IdColumnUtils.java
```diff
@@ -0,0 +1,67 @@
...
+   * @return JOOQ field corresponding to the valueGetters for the id columns of the entity type
+   */
+  public static Field<String[]> getResultIdValueGetter(EntityType entityType) {
+    List<Field<?>> idColumnValueGetters = new ArrayList<>();
+    getIdColumnValueGetters(entityType)
+      .forEach(valueGetterString -> idColumnValueGetters.add(field(valueGetterString)));
+    return DSL.cast(
+      DSL.array(idColumnValueGetters.toArray(new Field[0])),
+      String[].class
+    );
```
mweaver-ebsco, line 65: I wonder what this would look like as 1 long stream...
```java
    return List<Field<?>> idColumnValueGetters = getIdColumnValueGetters(entityType).stream()
      .map(DSL::field)
      .collect(Collectors.collectingAndThen(Collectors.toList(),
        idColumnValueGetters -> DSL.cast(
          DSL.array(idColumnValueGetters.toArray(new Field[0])),
          String[].class
        )
      );
```

Nope... That's definitely worse :D

mweaver-ebsco (approved):

ncovercash (approved):

src/main/java/org/folio/fqm/service/ResultSetService.java
bvsharp: The only way this would happen is if a record has a null value for one of its ID columns. So it could theoretically be a problem if we allow null values for ID columns, but I don't think it's something that would happen for our current entity types.

bvsharp: if we want to allow nullable id columns, we can change this to something like:
```
var columnValue = content.get(columnName);
keys.add(columnValue == null ? null : columnValue.toString());
```

src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -42,35 +46,70 @@ public class ResultSetRepository {
 
   public List<Map<String, Object>> getResultSet(UUID entityTypeId,
```
bvsharp, line 47: This method is kind of a mess, but it does drastically improve the performance while still allowing for composite result ids. I figured we can just get it in now and clean it up some later.

mweaver-ebsco: I wouldn't worry about it too much right now, but I bet just breaking out it into a couple of smaller methods would probably go a long way toward cleaning it up

mweaver-ebsco (approved):

ncovercash (approved):

PR #131, opened 2024-02-09
MODFQMMGR-134 Implement support for value source APIs

mweaver-ebsco:
This commit makes it so we can retrieve available values for columns from configured APIs. It also expands the functionality of the column values API endpoint by making it support values coming from the entity type definition, the DB, or APIs (prior to this, it only supported the DB)

### DIFF
--- a/src/main/java/org/folio/fqm/client/SimpleHttpClient.java
+++ b/src/main/java/org/folio/fqm/client/SimpleHttpClient.java
@@ -1,0 +1,18 @@
+package org.folio.fqm.client;
+
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Service;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+@Service
+@FeignClient(name = "simple-http-client", url = ".")
+public interface SimpleHttpClient {
+  /**
+   * Retrieve arbitrary data from a FOLIO API endpoint.
+   * @return the body of the response (JSON)
+   */
+  @GetMapping(value = "/{path}", produces = MediaType.APPLICATION_JSON_VALUE)
+  String get(@PathVariable String path);
+}
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -1,12 +1,17 @@
 package org.folio.fqm.service;
 
+import com.jayway.jsonpath.DocumentContext;
+import com.jayway.jsonpath.JsonPath;
 import lombok.RequiredArgsConstructor;
 
 import org.folio.fqm.domain.dto.EntityTypeSummary;
+import org.folio.fqm.client.SimpleHttpClient;
+import org.folio.fqm.exception.ColumnNotFoundException;
 import org.folio.fqm.exception.EntityTypeNotFoundException;
 import org.folio.fqm.repository.EntityTypeRepository;
 import org.folio.querytool.domain.dto.ColumnValues;
 import org.folio.querytool.domain.dto.EntityType;
+import org.folio.querytool.domain.dto.EntityTypeColumn;
 import org.folio.querytool.domain.dto.ValueWithLabel;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Service;
@@ -14,9 +19,11 @@ import org.springframework.transaction.annotation.Transactional;
 
 import static org.folio.fqm.repository.EntityTypeRepository.ID_FIELD_NAME;
 
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -24,11 +31,11 @@ import java.util.UUID;
 @Service
 @RequiredArgsConstructor
 public class EntityTypeService {
-  private static final String COLUMN_VALUE_SEARCH_FQL = "{\"%s\": {\"$regex\": \"%s\"}}";
   private static final int COLUMN_VALUE_DEFAULT_PAGE_SIZE = 1000;
   private final EntityTypeRepository entityTypeRepository;
   private final LocalizationService localizationService;
   private final QueryProcessorService queryService;
+  private final SimpleHttpClient columnValueClient;
 
   /**
    * Returns the list of all entity types.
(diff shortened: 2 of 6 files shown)


src/main/java/org/folio/fqm/client/SimpleHttpClient.java
```diff
@@ -0,0 +1,18 @@
+package org.folio.fqm.client;
+
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Service;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+@Service
+@FeignClient(name = "simple-http-client", url = ".")
```
mweaver-ebsco, line 10: It took me soooo long to figure this part out :P The name is required, but it sets the base path for the client. After a lot of trial and error, I figured out that setting `url` to `"."` will override the base path with something that doesn't actually do anything, so that the path in `get()` actually works

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -1,34 +1,41 @@
...
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 
 @Service
 @RequiredArgsConstructor
 public class EntityTypeService {
-  private static final String COLUMN_VALUE_SEARCH_FQL = "{\"%s\": {\"$regex\": \"%s\"}}";
```
mweaver-ebsco, line 27: I didn't need to change this, but it was driving me nuts that we had this string that was only used in one spot, but the var was defined really far from where it's used

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -70,7 +77,54 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
-    String fql = String.format(COLUMN_VALUE_SEARCH_FQL, columnName, searchText == null ? "" : searchText);
+    searchText = searchText == null ? "" : searchText;
+
+    EntityType entityType = entityTypeRepository.getEntityTypeDefinition(entityTypeId)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    EntityTypeColumn column = entityType.getColumns()
+      .stream()
+      .filter(col -> Objects.equals(columnName, col.getName()))
+      .findFirst()
+      .orElseThrow(() -> new ColumnNotFoundException(entityType.getName(), columnName));
```
mweaver-ebsco, line 88: This new stuff at the top of the method is what broke a bunch of tests. Before, it didn't matter if tests used a random UUID as the entity type ID. But now that it actually looks it up, there needs to be an `EntityType` that backs the UUID

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -70,7 +77,54 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
+      .filter(valueWithLabel -> valueWithLabel.getLabel().contains(searchText))
+      .toList();
+    return new ColumnValues().content(filteredColumns);
+  }
+
+  private ColumnValues getColumnValuesFromApi(EntityTypeColumn column, String searchText) {
+    String rawJson = columnValueClient.get(column.getValueSourceApi().getPath());
+    DocumentContext parsedJson = JsonPath.parse(rawJson);
+    List<String> values = parsedJson.read(column.getValueSourceApi().getValueJsonPath());
+    List<String> labels = parsedJson.read(column.getValueSourceApi().getLabelJsonPath());
```
mweaver-ebsco, line 113: JsonPath = awesome :)

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -70,7 +77,54 @@ public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId) {
...
+    List<ValueWithLabel> results = new ArrayList<>(values.size());
+    for (int i = 0; i < values.size(); i++) {
+      String value = values.get(i);
+      String label = labels.get(i);
+      if (label.contains(searchText)) {
+        results.add(new ValueWithLabel().value(value).label(label));
+      }
+    }
+    return new ColumnValues().content(results);
+  }
```
mweaver-ebsco, line 124: So in functional programming languages, there tends to be a function called `zip` that combines 2 collections together into a collection of pairs. Oracle actually had a `Stream.zip()` in the pre-releases of Java 8, but decided to remove it before releasing. If we still had that, then this would have been soooooo much cleaner as a stream... This would have looked something like
```java
return values.stream()
  .zip(labels.stream())
  .filter(valueAndLabel -> valueAndLabel.getSecond().contains(searchText))
  .map(valueAndLabel -> new ValueWithLabel().value(valueAndLabel.getFirst()).label(valueAndLabel.getSecond())
  .toList()
```

src/main/resources/db/changelog/changes/v1.1.0/update-drv-item-details-definition.xml
```diff
@@ -239,14 +239,14 @@
...
-                "entityTypeId": "917ea5c8-cafe-4fa6-a942-e2388a88c6f6"
-              },
               "dataType": {
                 "dataType": "stringType"
               },
-              "idColumnName": "item_material_type_id",
+              "valueSourceApi": {
+                "path": "material-types",
+                "valueJsonPath": "$.mtypes.*.id",
+                "labelJsonPath": "$.mtypes.*.name"
```
mweaver-ebsco, line 248: Alternatively, this could jave been
```json
                "valueJsonPath": "$..id",
                "labelJsonPath": "$..name"
```
I opted for the more targetted json path, just to be safe

src/test/java/org/folio/fqm/service/EntityTypeServiceTest.java
```diff
@@ -136,13 +156,87 @@ void shouldFilterBySearchText() {
...
+    String valueColumnName = "column_name";
+    EntityType entityType = new EntityType()
+      .id(entityTypeId.toString())
+      .name("the entity type")
+      .columns(List.of(new EntityTypeColumn()
+        .name(valueColumnName)
+        .valueSourceApi(new ValueSourceApi()
+          .path("fake-path")
+          .valueJsonPath("$.what.ever.dude.*.theValue") // Approach 1: Explicitly use the full path
+          .labelJsonPath("$..theLabel") // Approach 2: Just dive all the way down and find everything with this key
```
mweaver-ebsco, line 203: It's a little hidden, but I added the comments here to illustrate the different ways to do the JsonPath, since it seemed worth noting _somewhere_

ncovercash (approved):

bvsharp (approved):

kjain110 (approved):

src/test/java/org/folio/fqm/service/EntityTypeServiceTest.java
```diff
@@ -136,13 +156,87 @@ void shouldFilterBySearchText() {
...
+        )
+      ));
+
+    when(repo.getEntityTypeDefinition(entityTypeId)).thenReturn(Optional.of(entityType));
+    when(simpleHttpClient.get("fake-path")).thenReturn("""
+           {
+             "what": {
```
kjain110, line 210: Nice :)

PR #138, opened 2024-02-14
Handle case insensitivity based on column datatype

bvsharp:
## Purpose
Handle case insensitivity based on column datatype instead of the type of the condition value (this sounds terrible). Now, FQM can handle integer/number types even if they are sent as strings.

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -44,6 +44,10 @@ public class FqlToSqlConverterService {
    * an exact comparison will be performed.
    */
   private static final String DATE_REGEX = "^\\d{4}-\\d{2}-\\d{2}$";
+  private static final String STRING_TYPE = "stringType";
+  private static final String RANGED_UUID_TYPE = "rangedUUIDType";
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -86,8 +90,9 @@ private static Condition handleEquals(EqualsCondition equalsCondition, EntityTyp
     if (isDateCondition(equalsCondition, entityType)) {
       return handleDate(equalsCondition, field);
     }
-    if (equalsCondition.value() instanceof String value) {
-      return caseInsensitiveComparison(equalsCondition, entityType, field, value, Field::equalIgnoreCase, Field::eq);
+    var dataType = getColumnDataType(entityType, equalsCondition);
```
ncovercash, line 93: Nit: var is stinky :(

ncovercash (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
bvsharp: totally agree, good eye

PR #139, opened 2024-02-15
MODFQMMGR-134 Fix a bug with the material type config

mweaver-ebsco:
This adds the `source` property back to item_material_type in the Items entity type, so that the query builder plugin knows how to ask for the available values.
This commit also adds "inventory-storage.material-types.collection.get" to the GET /entity-types sub-permissions, so that users who can access the entity types can also access the available values for that entity type

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.1.0/update-drv-item-details-definition.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/update-drv-item-details-definition.xml
@@ -248,6 +248,10 @@
                 "valueJsonPath": "$.mtypes.*.id",
                 "labelJsonPath": "$.mtypes.*.name"
               },
+              "source": {
+                "columnName": "item_material_type",
+                "entityTypeId": "0cb79a4c-f7eb-4941-a104-745224ae0292"
+              },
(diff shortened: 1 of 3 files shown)


src/main/resources/db/changelog/changes/v1.1.0/update-drv-item-details-definition.xml
```diff
@@ -248,6 +248,10 @@
                 "valueJsonPath": "$.mtypes.*.id",
                 "labelJsonPath": "$.mtypes.*.name"
               },
+              "source": {
+                "columnName": "item_material_type",
+                "entityTypeId": "0cb79a4c-f7eb-4941-a104-745224ae0292"
+              },
```
mweaver-ebsco, line 254: The query builder uses the property to configure how it requests the available values, so we actually still need it even if the values aren't coming from the DB. In this case, I set it up to just ask for the values for this column for the Items entity type, but an alternative approach could be to use the existing material types entity type and modify that one so that the values come from an API instead of the DB. I opted to do it all in this entity type because that will allow us to kill off an entity type which just feels good

ncovercash (approved):
looks like CI is mad because you haven't merged master in (new version of lib-fqm-query-processor)

kjain110 (approved):

kjain110:
CI is always mad

PR #140, opened 2024-02-16
Add datatype string to boolean custom fields

bvsharp:
## Purpose
Add datatype string to custom field columns so that they work again.

### DIFF
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -122,7 +122,7 @@ public class EntityTypeRepository {
 
     return new EntityTypeColumn()
       .name(value)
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -122,7 +122,7 @@ private EntityTypeColumn handleSingleCheckBox(String value, String refId, String
 
     return new EntityTypeColumn()
       .name(value)
-      .dataType(new BooleanType())
+      .dataType(new BooleanType().dataType("booleanType"))
```
mweaver-ebsco, line 125: Do we have this string defined as a constant somewhere that we can reuse, so that we don't have to put a hard-coded string here?

PR #142, opened 2024-02-19
[FQTM-6] Support fully qualified localized names for nested objects

ncovercash:
# [Jira FQTM-6](https://folio-org.atlassian.net/browse/FQTM-6)

## Purpose
This adds localization support for the nested property `labelAliasFullyQualified`, completing the last step in a fully localized definition of nested properties.

For example, fund distributions:
- Parent `fund_distributions` column: `labelAlias=Fund distribution`
- Child `amount` property: `labelAlias=Amount` (for showing within results)
- Child `amount` property: `labelAliasFullyQualified=Fund distribution — amount` or `Fund distribution amount`, etc.


bvsharp (approved):

src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -23,7 +23,9 @@ public class LocalizationService {
   private static final String ENTITY_TYPE_COLUMN_LABEL_TRANSLATION_TEMPLATE = "mod-fqm-manager.entityType.%s.%s";
   // refers to a property inside an objectType column inside the entity type, e.g. "City" inside "Address" column inside "Users"
   private static final String ENTITY_TYPE_COLUMN_NESTED_LABEL_TRANSLATION_TEMPLATE =
-    "mod-fqm-manager.entityType.%s.%s.%s";
+  "mod-fqm-manager.entityType.%s.%s.%s";
```
mweaver-ebsco, line 26: Nit: It looks like the line above your change got its formatting a little mangled (it's no longer indented as much as the other values nearby). That could easily just be GitHub being GitHub, though

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/LocalizationService.java
ncovercash: Oops yep, sometimes copy/pasting a line does this in my editor. Thanks!

PR #144, opened 2024-02-20
[MODFQMMGR-151] Support nested object queries

ncovercash:
# [Jira MODFQMMGR-151](https://folio-org.atlassian.net/browse/MODFQMMGR-151)

## Purpose
Allow queries on nested object properties, e.g. querying for entity type `Purchase Order Lines` where a `fund_distribution` `$contains` a `code` of `ASIAHIST`.

This allows queries such as:

```json
{"fund_distribution[*]->code": {"$contains": "AFRICAHIST"}}
```

## Approach
This is done by abstracting methods which relied on `EntityTypeColumn` to use the new superclass `Field` introduced in https://github.com/folio-org/folio-query-tool-metadata/pull/42


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -1,15 +1,27 @@
...
 import org.folio.fql.service.FqlService;
-import org.folio.fqm.exception.ColumnNotFoundException;
+import org.folio.fql.service.FqlValidationService;
+import org.folio.fqm.exception.FieldNotFoundException;
 import org.folio.fqm.utils.SqlFieldIdentificationUtils;
 import org.folio.querytool.domain.dto.DateType;
 import org.folio.querytool.domain.dto.EntityDataType;
 import org.folio.querytool.domain.dto.EntityType;
-import org.folio.querytool.domain.dto.EntityTypeColumn;
+import org.folio.querytool.domain.dto.Field;
```
ncovercash, line 23: I decided that it'd be clearer to use our Field as `Field` rather than Jooq's, since we refer to our `EntityType` and other objects directly.  This makes it a little more verbose to say when we're manipulating/returning a Jooq field, instead

src/main/java/org/folio/fqm/resource/EntityTypeController.java
```diff
@@ -34,8 +34,8 @@ public ResponseEntity<List<EntityTypeSummary>> getEntityTypeSummary(List<UUID> e
   }
 
   @Override
-  public ResponseEntity<ColumnValues> getColumnValues(UUID entityTypeId, String columnName, String search) {
-    return ResponseEntity.ok(entityTypeService.getColumnValues(entityTypeId, columnName, search));
+  public ResponseEntity<ColumnValues> getColumnValues(UUID entityTypeId, String fieldName, String search) {
```
ncovercash, line 37: Ideally, I'd like to rename this type to `FieldValues` and the method to `getFieldValues`, but that needs to come from folio-query-tool-metadata and will come later.

mweaver-ebsco: Unfortunately, this API endpoint actually has "columns" in the path :/ I'm all for changing that, but it will be a pain (my vote would be to add a new endpoint that uses "fields" instead of "columns" and keep the old one around, then mark the old one as deprecated and have it just call the new one, then we can update the query builder at our leisure and remove the deprecated endpoint later)

ncovercash:
Idk why Sonar is mad at these...

mweaver-ebsco (approved):

bvsharp (approved):

PR #146, opened 2024-02-21
MODFQMMGR-167: Fix fields that cannot be queried

bvsharp:
## Purpose
[MODFQMMGR-167](https://folio-org.atlassian.net/browse/MODFQMMGR-167): Fix a few fields that can't be queried

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
@@ -31,7 +31,7 @@
                   "dataType": "stringType"
                 }
               },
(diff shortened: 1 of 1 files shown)


mweaver-ebsco (approved):

src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -187,6 +187,7 @@
                 }
               },
               "valueGetter": "ARRAY(SELECT jsonb_array_elements_text(purchase_order.jsonb -> 'notes'))",
+              "filterValueGetter": "ARRAY(SELECT lower(jsonb_array_elements_text(purchase_order.jsonb -> 'notes')))",
```
mweaver-ebsco, line 190: Since there's a filterValueGetter now, do we also need a valueFunction?

bvsharp: The valueFunction would just be a lower(), but this is handled in java code for the handleContains/handleNotContains methods (because array fields don't play nicely with valueFunctions right now). While working on the new array operator tickets, I'm going to look into how easy it would be to remove that logic from handleContains/handleNotContains, in which case we will need a lower() valueFunction here.

PR #151, opened 2024-02-27
[MODFQMMGR-180] Use proper checks for insensitive comparison

ncovercash:
# [Jira MODFQMMGR-180](https://folio-org.atlassian.net/browse/MODFQMMGR-180)

## Purpose
As part of #147 , we added some logic to perform filtering on underlying ID columns, rather than the derived columns (e.g. vendor_id instead of vendor_code). This needs to be updated to also check the type of the underlying column, to ensure that we appropriately apply lowercase, date, and other conversions.


ncovercash:
I'm not sure if we should do the same thing for this method's `getField`? I'd assume yes, but I'm not as sure as to how this one is used.
image

kjain110 (approved):

mweaver-ebsco (approved):

PR #162, opened 2024-03-05
MODFQMMGR-147:adding more fields

kjain110:


bvsharp (approved):

translations/mod-fqm-manager/en.json
```diff
@@ -335,5 +335,10 @@
   "entityType.drv_organization_details.expected_activation_interval": "Organization expected activation interval",
   "entityType.drv_organization_details.vendor_currencies": "Organization vendor currencies",
   "entityType.drv_organization_details.tax_id": "Organization tax ID",
-  "entityType.drv_organization_details.tax_percentage": "Organization tax percentage"
+  "entityType.drv_organization_details.tax_percentage": "Organization tax percentage",
+  "entityType.drv_organization_details.agreements": "Organization agreements",
+  "entityType.drv_organization_details.agreements.name": "Organization vendor terms name",
```
ncovercash, line 340: There are two translations for each nested property. If you check the logs of mod-fqm-manager while you're querying against it, you should see warnings for missing translations.
```suggestion
  "entityType.drv_organization_details.agreements.name": "Vendor terms name",
  "entityType.drv_organization_details.agreements.name._qualified": "Organization vendor terms name",
```

Here's how they differ:
- The normal `.name` is used as the inner column label, and likely needs the parent label to make sense (e.g. "Notes" doesn't make sense alone unless it's inside "Vendor terms"):
image
- The `._qualified` is shown in the query builder/etc and should make sense alone, e.g. "Vendor terms notes"

translations/mod-fqm-manager/en.json
```diff
@@ -335,5 +335,14 @@
...
+  "entityType.drv_organization_details.tax_percentage": "Organization tax percentage",
+  "entityType.drv_organization_details.agreements": "Organization agreements",
+  "entityType.drv_organization_details.agreements.name": "Vendor terms name",
+  "entityType.drv_organization_details.agreements.name._qualified": "Organization vendor terms name",
+  "entityType.drv_organization_details.agreements.discount": "Discount percentage",
+  "entityType.drv_organization_details.agreements.discount._qualified": "Organization discount percentage",
+  "entityType.drv_organization_details.agreements.referenceUrl": "Reference URL",
+  "entityType.drv_organization_details.agreements.referenceUrl._qualified": "Organization reference URL",
+  "entityType.drv_organization_details.agreements.notes": "Notes",
+  "entityType.drv_organization_details.agreements.notes._qualified": "Organization notes"
```
ncovercash, line 347: Would it be organization? Or something like "Vendor terms notes" instead? What describes this column on its own

kjain110: its inside agreements array

ncovercash: then maybe "Agreements notes"?

ncovercash (approved):

PR #167, opened 2024-03-06
[MODFQMMGR-185] Add categories, acquisition units to organization contact entity type

ncovercash:
# [Jira MODFQMMGR-185](https://folio-org.atlassian.net/browse/MODFQMMGR-185)

### DIFF
--- a/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
@@ -293,6 +293,39 @@
               },
               "valueGetter": "org.jsonb ->> 'urls'",
               "visibleByDefault": false
+            },
+            {
+              "name": "acquisition_unit_id",
+              "dataType": {
+                "dataType": "arrayType",
+                "itemDataType": {
+                  "dataType": "rangedUUIDType"
+                }
+              },
+              "queryable": false,
+              "valueGetter": "( SELECT array_agg(acq_id.value::text) FILTER (WHERE (acq_id.value::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text(org.jsonb -> 'acqUnitIds'::text) acq_id(value))",
+              "filterValueGetter": "( SELECT array_agg(lower(acq_id.value::text)) FILTER (WHERE (acq_id.value::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text(org.jsonb -> 'acqUnitIds'::text) acq_id(value))",
+              "valueFunction": "lower(:value)",
+              "visibleByDefault": false
+            },
+            {
+              "name": "acquisition_unit_name",
+              "dataType": {
+                "dataType": "arrayType",
+                "itemDataType": {
+                  "dataType": "stringType"
+                }
+              },
+              "queryable": true,
+              "valueGetter": "( SELECT array_agg(acq_unit.jsonb ->> 'name'::text) FILTER (WHERE (acq_unit.jsonb ->> 'name'::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text((org.jsonb -> 'acqUnitIds'::text)) record(value) JOIN src_acquisitions_unit acq_unit ON lower(record.value::text) = acq_unit.id::text)",
+              "filterValueGetter": "( SELECT array_agg(lower(acq_unit.jsonb ->> 'name'::text)) FILTER (WHERE (acq_unit.jsonb ->> 'name'::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text((org.jsonb -> 'acqUnitIds'::text)) record(value) JOIN src_acquisitions_unit acq_unit ON (record.value::text) = acq_unit.id::text)",
+              "valueFunction": "lower(:value)",
+              "visibleByDefault": false,
+              "idColumnName": "acquisition_unit_name",
+              "source": {
+                "entityTypeId": "cc51f042-03e2-43d1-b1d6-11aa6a39bc78",
+                "columnName": "acquisitions_name"
+              }
             }
           ],
           "defaultSort": [
--- a/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
@@ -2,6 +2,16 @@
                    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+  <changeSet id="[MODFQMMGR-185] Create supporting organization contact categories view" author="novercash@ebsco.com">
+    <preConditions onFail="CONTINUE">
+      <tableExists tableName = "organization_types" schemaName="${tenant_id}_mod_organizations_storage"/>
+    </preConditions>
+    <createView
+      replaceIfExists="true"
+      viewName="src_organization_categories">
+      SELECT * FROM ${tenant_id}_mod_organizations_storage.categories
+    </createView>
+  </changeSet>
   <changeSet id="[MODFQMMGR-176] Create entity type for organizations contact info" author="novercash@ebsco.com" runOnChange="true">
     <!-- Insert the row, then fill it in later, for readability. -->
     <sql>
--- a/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
+++ b/src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
@@ -158,7 +158,6 @@
                       "property": "addressLine1",
                       "dataType": { "dataType": "stringType" },
                       "queryable": true,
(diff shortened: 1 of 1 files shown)


src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -158,7 +168,6 @@
                       "property": "addressLine1",
                       "dataType": { "dataType": "stringType" },
                       "queryable": true,
-                      "valueGetter": "( SELECT array_agg(elems.value ->> 'addressLine1') FROM jsonb_array_elements(org.jsonb->'addresses') AS elems)",
```
ncovercash, line 161: We don't retrieve values of these inner columns (only the outer one, which gives back a JSON blob), so these serve zero purpose.

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -296,15 +295,67 @@
...
+                    },
+                    {
+                      "name": "category_ids",
+                      "property": "categories",
+                      "dataType": {
+                        "dataType": "arrayType",
+                        "itemDataType": { "dataType": "rangedUUIDType" }
+                      },
+                      "queryable": false,
+                      "filterValueGetter": "( SELECT array_agg(lower(elems#>>'{}')) FROM jsonb_array_elements(org.jsonb->'urls') AS url_obj, jsonb_array_elements(url_obj->'categories') as elems )"
```
ncovercash, line 309: The `#>>'{}'` is the only way I've found to turn a JSON string `"foo"` into the string `foo` with no quotes/etc. Normally, we could use `->>` in lieu of `->`, but since `->categories` returns an array of strings, and `jsonb_array_elements` iterates over each as `jsonb`, we can't do that.

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -296,15 +295,67 @@
...
+                    },
+                    {
+                      "name": "category_names",
+                      "property": "categoriesNames",
+                      "dataType": {
+                        "dataType": "arrayType",
+                        "itemDataType": { "dataType": "stringType" }
+                      },
+                      "queryable": true,
```
ncovercash, line 318: I don't think our existing ID column work supports nested things like this, especially this deeply nested. So I'm thinking we should leave as just having "names" queryable? We lose some resiliency if the underlying category is renamed, but otherwise it should be fine.

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -296,15 +295,67 @@
...
+                      "filterValueGetter": "( SELECT array_agg(lower(cats.jsonb->>'value')) FROM jsonb_array_elements(org.jsonb->'urls') as url, jsonb_array_elements(url->'categories') as category_id JOIN src_organization_categories cats ON cats.id = (category_id#>>'{}')::uuid )"
```
ncovercash, line 319: This does a bit of magic that's hard to read slammed into one line.

```sql
  -- 4. get the name of the category, lowercase, and return
  SELECT array_agg(lower(cats.jsonb->>'value'))
  FROM jsonb_array_elements(org.jsonb->'urls') as url, -- 1. unpacks urls so we can look at each url object, one at a time
    jsonb_array_elements(url->'categories') as category_id -- 2. gets each category for this url
    JOIN src_organization_categories cats ON cats.id = (category_id#>>'{}')::uuid -- 3. grabs corresponding category information from view
```

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -296,15 +295,67 @@
...
                     }
                   ]
                 }
               },
-              "valueGetter": "org.jsonb ->> 'urls'",
+              "valueGetter": "( SELECT array_agg(url || jsonb_build_object('categoryNames', ( SELECT array_agg(cats.jsonb->>'value') FROM jsonb_array_elements(url->'categories') as category_id JOIN src_organization_categories cats on cats.id = (category_id#>>'{}')::uuid ))) FROM jsonb_array_elements(org.jsonb->'urls') as url)",
```
ncovercash, line 324: This also does a little magic. Category names are not present in the original `jsonb` (only IDs), so we need to add them in.  In essence, we unroll the normal `urls` and apply similar logic as the `filterValueGetter` for the category names.

```sql

  SELECT array_agg(
      url || jsonb_build_object( -- || will merge two json objects, allowing us to add a categoryNames property to each
        'categoryNames',
        (
          SELECT array_agg(cats.jsonb->>'value')
          FROM jsonb_array_elements(url->'categories') as category_id
            JOIN src_organization_categories cats on cats.id = (category_id#>>'{}')::uuid
        )
      )
    )
  FROM jsonb_array_elements(org.jsonb->'urls') as url
```

ncovercash: Result: we get category names here, too!
image

ncovercash: oops 
image

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -2,6 +2,16 @@
                    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+  <changeSet id="[MODFQMMGR-185] Create supporting organization contact categories view" author="novercash@ebsco.com">
+    <preConditions onFail="CONTINUE">
+      <tableExists tableName = "organization_types" schemaName="${tenant_id}_mod_organizations_storage"/>
```
kjain110, line 7: you need to change the tableName to categories

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
ncovercash: fixed nesting issue. top comment still is good tho

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -203,14 +208,33 @@
                       "property": "country",
                       "dataType": { "dataType": "stringType" },
                       "queryable": true,
-                      "valueGetter": "( SELECT array_agg(elems.value ->> 'country') FROM jsonb_array_elements(org.jsonb->'addresses') AS elems)",
                       "filterValueGetter": "( SELECT array_agg(lower(elems.value ->> 'country')) FROM jsonb_array_elements(org.jsonb -> 'addresses') AS elems)",
                       "valueFunction": "lower(:value)"
+                    },
```
ncovercash, line 213: scroll down to see an explanation of this

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
bvsharp: yeah, I don't think it would be worth it to rework all the other logic just for one field, so I like this idea

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
```diff
@@ -296,15 +355,67 @@
...
+                "itemDataType": {
+                  "dataType": "stringType"
+                }
+              },
+              "queryable": true,
+              "valueGetter": "( SELECT array_agg(acq_unit.jsonb ->> 'name'::text) FILTER (WHERE (acq_unit.jsonb ->> 'name'::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text((org.jsonb -> 'acqUnitIds'::text)) record(value) JOIN src_acquisitions_unit acq_unit ON lower(record.value::text) = acq_unit.id::text)",
+              "filterValueGetter": "( SELECT array_agg(lower(acq_unit.jsonb ->> 'name'::text)) FILTER (WHERE (acq_unit.jsonb ->> 'name'::text) IS NOT NULL) AS array_agg FROM jsonb_array_elements_text((org.jsonb -> 'acqUnitIds'::text)) record(value) JOIN src_acquisitions_unit acq_unit ON (record.value::text) = acq_unit.id::text)",
+              "valueFunction": "lower(:value)",
+              "visibleByDefault": false,
+              "idColumnName": "acquisition_unit_name",
```
bvsharp, line 414: should this be acquisition_unit_id?

bvsharp (approved):

src/main/resources/db/changelog/changes/v1.1.0/insert-organizations-contact-info-entity-type.xml
kjain110: it should be` "idColumnName": "acqunit_ids"`

kjain110 (approved):

PR #168, opened 2024-03-07
MODFQMMGR-186 Bump Spring Boot dependency version

mweaver-ebsco:
This commit bumps our dependency on Spring Boot from 3.1.2 to 3.2.3 (the latest stable release). This also bumps folio-spring-base from 7.1.2 to 8.1.0, to keep our transitive Spring dependencies in sync with our Boot version

### DIFF
--- a/src/test/java/org/folio/fqm/ModFqmManagerApplicationTest.java
+++ b/src/test/java/org/folio/fqm/ModFqmManagerApplicationTest.java
@@ -2,8 +2,7 @@ package org.folio.fqm;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import javax.validation.Valid;
-
+import jakarta.validation.Valid;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/fqm/ModFqmManagerApplicationTest.java
```diff
@@ -2,8 +2,7 @@
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import javax.validation.Valid;
-
+import jakarta.validation.Valid;
```
ncovercash, line 5: called it, lol (a pointless javax -> jakarta)

ncovercash (approved):

bvsharp (approved):

src/test/java/org/folio/fqm/ModFqmManagerApplicationTest.java
mweaver-ebsco: I immediately thought of your comment when this one came up 😄

PR #171, opened 2024-03-07
MODFQMMGR-188: fix null/empty operator for nested fields

bvsharp:
## Purpose
[MODFQMMGR-188](https://folio-org.atlassian.net/browse/MODFQMMGR-188): fix null/empty operator for nested fields

For an array field _arrayField_ and a nested element _nestedElement_, $empty queries are handled in the following ways:

- If arrayField is not present, then 'arrayField.nestedElement IS EMPTY' returns true
-  If arrayField is an empty array, then 'arrayField.nestedElement IS EMPTY' returns true
- If arrayField is not empty, but none of the elements contain nestedElement, then 'arrayField.nestedElement IS EMPTY' returns true
- If arrayField is not empty, and any of the elements contain nestedElement, then 'arrayField.nestedElement IS EMPTY' returns false

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -65,6 +65,31 @@ public class FqlToSqlConverterService {
   private static final String RANGED_UUID_TYPE = "rangedUUIDType";
   private static final String OPEN_UUID_TYPE = "openUUIDType";
   private static final String DATE_TYPE = "dateType";
+  public static final String ALL_NULLS = """
+          true = ALL(
+            SELECT(
+              unnest(
+                cast(
+                  %s
+                  as varchar[]
+                )
+              )
+            ) IS NULL
+          )
+          """;
+  public static final String NOT_ALL_NULLS = """
+          false = ANY(
+            SELECT(
+              unnest(
+                cast(
+                  %s
+                  as varchar[]
+                )
+              )
+            ) IS NULL
+          )
+          """;
+
 
   private final FqlService fqlService;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -65,6 +65,31 @@ public class FqlToSqlConverterService {
...
+            SELECT(
+              unnest(
+                cast(
+                  %s
+                  as varchar[]
+                )
+              )
+            ) IS NULL
+          )
+          """;
```
bvsharp, line 79: I spent too long trying to get this to work as a JOOQ statement, before eventually just settling on this beautiful monstrosity.

bvsharp, line 79: I spent too long trying to get this to work as a JOOQ statement, before eventually just settling on this beautiful monstrosity.

kjain110 (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -271,16 +296,19 @@ private static Condition handleNotContainsAny(NotContainsAnyCondition notContain
...
     var nullCondition = isEmpty ? field.isNull() : field.isNotNull();
+
     return switch (fieldType) {
       case STRING_TYPE -> isEmpty ? nullCondition.or(field.eq("")) : nullCondition.and(field.ne(""));
       case "arrayType" -> {
-        var cardinality = cardinality(DSL.cast(field, String[].class));
-        yield isEmpty ? nullCondition.or(cardinality.eq(0)) :  nullCondition.and(cardinality.ne(0));
+        var cardinality = cardinality(cast(field, String[].class));
+        if (isEmpty) {
+          yield nullCondition.or(cardinality.eq(0)).or(ALL_NULLS.formatted(field));
```
mweaver-ebsco, line 308: Rather than pass values in with `.formatted()`, can we parameterize this? If not, we risk opening up a pretty big SQL injection hole

ncovercash (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
mweaver-ebsco: I guess it's already partially parameterized by `or()`, so there's probably not really a security concern here

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
bvsharp: I totally understand the concern, that's the main reason I didn't really like this approach. Since the entire condition is being passed as a string, I couldn't figure out how to parameterize it in this context (since we want to pass the whole argument as a string, but just want the field name to be parameterized). But like you said, I think it should be ok with the `or()`, plus the fact that only our predefined valueGetters/filterValueGetters will ever be used here.

PR #182, opened 2024-03-13
Ensure id columns are always included in query contents

bvsharp:
## Purpose
Ensure id columns are always included in query contents. This change is needed to make sure the GET /query/contents API works whether the fields parameter:
  - is not included
  - is included
  - is included, but is missing id columns

### DIFF
--- a/src/main/java/org/folio/fqm/service/QueryManagementService.java
+++ b/src/main/java/org/folio/fqm/service/QueryManagementService.java
@@ -193,23 +193,20 @@ public class QueryManagementService {
   }
 
   public List<Map<String, Object>> getContents(UUID entityTypeId, List<String> fields, List<List<String>> ids) {
+    entityTypeService
+      .getEntityTypeDefinition(entityTypeId)
+      .map(IdColumnUtils::getIdColumnNames)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId))
+      .forEach(colName -> {
+        if (!fields.contains(colName)) {
+          fields.add(colName);
+        }
+      });
     return resultSetService.getResultSet(entityTypeId, fields, ids);
   }
 
   private List<Map<String, Object>> getContents(UUID queryId, UUID entityTypeId, List<String> fields, boolean includeResults, int offset, int limit) {
     if (includeResults) {
-      if (CollectionUtils.isEmpty(fields)) {
-        Query query = queryRepository.getQuery(queryId, false)
-          .orElseThrow(() -> new QueryNotFoundException(queryId));
-        Fql fql = fqlService.getFql(query.fqlQuery());
-        fields = fqlService.getFqlFields(fql)
-          .stream()
-          .map(FqlField::getColumnName)
-          .collect(Collectors.toList());
-      }
-      if (!fields.contains("id")) {
-        fields.add("id");
-      }
       List<List<String>> resultIds = queryResultsRepository.getQueryResultIds(queryId, offset, limit);
       return resultSetService.getResultSet(entityTypeId, fields, resultIds);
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/fqm/service/QueryManagementService.java
```diff
@@ -193,6 +193,15 @@ public List<List<String>> getSortedIds(UUID queryId, int offset, int limit) {
...
   public List<Map<String, Object>> getContents(UUID entityTypeId, List<String> fields, List<List<String>> ids) {
+    entityTypeService
+      .getEntityTypeDefinition(entityTypeId)
+      .map(IdColumnUtils::getIdColumnNames)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId))
+      .forEach(colName -> {
+        if (!fields.contains(colName)) {
+          fields.add(colName);
+        }
+      });
```
bvsharp, line 204: We didn't previously need this, since we ensured that any query or list that was saved included id columns in its list of fields. But now that we're allowing a subset of columns to be selected (and in the future, only these columns will be sent from the UI to the BE), we need to make sure that the id columns are always included in the results, whether they're included in the request or not.

src/main/java/org/folio/fqm/service/QueryManagementService.java
```diff
@@ -207,8 +216,14 @@ private List<Map<String, Object>> getContents(UUID queryId, UUID entityTypeId, L
...
-      if (!fields.contains("id")) {
-        fields.add("id");
+      List<String> idColumnNames = entityTypeService
+        .getEntityTypeDefinition(entityTypeId)
+        .map(IdColumnUtils::getIdColumnNames)
+        .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+      for (String idColumnName : idColumnNames) {
+        if (!fields.contains(idColumnName)) {
+          fields.add(idColumnName);
+        }
```
bvsharp, line 226: This is a holdover from the composite id ticket. I missed it because I wasn't thinking about the synchronous GET /query API.

src/test/java/org/folio/fqm/service/QueryManagementServiceTest.java
```diff
@@ -242,106 +242,6 @@ void shouldReturnQueryDetailsWithContents() {
     assertEquals(expectedDetails, actualDetails);
   }
 
-  @Test
-  void shouldReturnQueryContentsWithOnlyIdsForEmptyFields() {
```
bvsharp, line 246: This test (and all the other removed ones) were used to ensure that id column logic was handled correctly even if idColumns weren't in the query's list of fields. But since we're now checking that during the POST, these are no longer needed.

src/main/java/org/folio/fqm/service/QueryManagementService.java
```diff
@@ -193,23 +193,20 @@ public List<List<String>> getSortedIds(UUID queryId, int offset, int limit) {
...
-          .orElseThrow(() -> new QueryNotFoundException(queryId));
-        Fql fql = fqlService.getFql(query.fqlQuery());
-        fields = fqlService.getFqlFields(fql)
-          .stream()
-          .map(FqlField::getColumnName)
-          .collect(Collectors.toList());
-      }
-      if (!fields.contains("id")) {
-        fields.add("id");
-      }
```
bvsharp, line 212: This is in the getContents method that is used to retrieve contents for a query. The POST query API now performs does this check, so the fields for a query will never be empty, making this redundant.

kjain110 (approved):

ncovercash (approved):

PR #191, opened 2024-03-15
Display all currencies in POL currency dropdown

bvsharp:
## Purpose
Display all currencies in pol_currency dropdown


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -96,6 +97,10 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
       return getFieldValuesFromApi(field, searchText);
     }
 
+    if (field.getName().equals("pol_currency")) {
```
bvsharp, line 100: Kind of gross to hardcode this, but the alternative would be to add some kind of "isCurrencyField" to the entity type column definition, which seems even worse.

ncovercash (approved):

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
mweaver-ebsco: Probably not worth doing now, but eventually, we could maybe add something more generic than "isCurrencyField" ("hasCustomAvailableValueLogic" or something? I dunno, that's pretty terrible... We should probably refactor the entity type class at some point, so that the different ways to retrieve available values all live under 1 key, rather than having several top-level keys), then in a separate method/class put custom handlers for specific columns

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
bvsharp: Yeah, I like this idea (for the future)

kjain110 (approved):

PR #195, opened 2024-03-19
Sort currencies alphabetically

bvsharp:
## Purpose
Sort currencies alphabetically in dropdown menu.

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -18,6 +18,7 @@ import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import static java.util.Comparator.comparing;
 import static org.folio.fqm.repository.EntityTypeRepository.ID_FIELD_NAME;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
```diff
@@ -333,6 +333,10 @@
               "valueGetter": "src_purchase_order_line.jsonb -> 'cost' ->> 'currency'",
               "filterValueGetter": "lower(src_purchase_order_line.jsonb -> 'cost' ->> 'currency')",
               "valueFunction": "lower(:value)",
+              "source": {
+                "entityTypeId": "90403847-8c47-4f58-b117-9a807b052808",
+                "columnName": "pol_currency"
+              },
```
bvsharp, line 339: This is kind of meaningless, but I had to add it in so that the query builder sends the request to GET /columns/pol_currency/values. For it to call that API, it's expecting either a "values", "source" or a "valueSourceApi" field here. POL currency doesn't actually use any of these, but we still need one present to get the call to happen.

kjain110 (approved):

src/main/resources/db/changelog/changes/v1.1.0/update-drv-purchase-order-line-details-definition.xml
mweaver-ebsco: Yeah, I've run into this, too :/ It's a little gross, but also makes a small amount of sense (it tells the UI how to ask for the values). I'd definitely prefer that there only be 1 way for it to get the values, though, rather than have it depend on `source`

ncovercash (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -149,13 +149,15 @@ private ColumnValues getFieldValuesFromEntityType(UUID entityTypeId, String fiel
...
-      Currency
+      new ArrayList<>(Currency
         .getAvailableCurrencies()
         .stream()
         .map(currency -> new ValueWithLabel()
           .value(currency.getCurrencyCode())
           .label(String.format("%s (%s)", currency.getDisplayName(), currency.getCurrencyCode())))
-        .toList();
+        .toList());
+    Comparator<ValueWithLabel> comparator = (v1, v2) -> v1.getLabel().compareTo(v2.getLabel());
```
mweaver-ebsco, line 159: `Comparator.comparing(ValueWithLabel::getLabel)`

PR #196, opened 2024-03-19
MODFQMMGR-210 Split DB migrations into a fast synchronous stage and a slow asynchronous stage

mweaver-ebsco:
The second run will run a special "slow" context in a separate thread,
so that FQM can continue bootstrapping its DB without holding up other
modules

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqmTenantService.java
+++ b/src/main/java/org/folio/fqm/service/FqmTenantService.java
@@ -20,6 +20,15 @@ public class FqmTenantService extends TenantService {
   @Override
   public synchronized void createOrUpdateTenant(TenantAttributes tenantAttributes) {
     this.folioSpringLiquibase.setChangeLogParameters(Map.of("tenant_id", this.context.getTenantId()));
+
+    // Run all of the regular DB migration scripts (those that aren't marked as "slow")
+    this.folioSpringLiquibase.setContexts("!slow");
     super.createOrUpdateTenant(tenantAttributes);
+
+    // Run all of the slow DB migration scripts in a separate thread
+    this.folioSpringLiquibase.setContexts("slow");
+    var slowRun = new Thread(() -> super.createOrUpdateTenant(tenantAttributes));
+    slowRun.setDaemon(true);
+    slowRun.start();
   }
 }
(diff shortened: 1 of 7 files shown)


ncovercash (approved):

src/main/java/org/folio/fqm/service/FqmTenantService.java
```diff
@@ -20,6 +20,15 @@ public FqmTenantService(JdbcTemplate jdbcTemplate, FolioExecutionContext context
...
     this.folioSpringLiquibase.setChangeLogParameters(Map.of("tenant_id", this.context.getTenantId()));
+
+    // Run all of the regular DB migration scripts (those that aren't marked as "slow")
+    this.folioSpringLiquibase.setContexts("!slow");
     super.createOrUpdateTenant(tenantAttributes);
+
+    // Run all of the slow DB migration scripts in a separate thread
+    this.folioSpringLiquibase.setContexts("slow");
+    var slowRun = new Thread(() -> super.createOrUpdateTenant(tenantAttributes));
+    slowRun.setDaemon(true);
```
mweaver-ebsco, line 31: Just to be safe, I set it as a daemon thread, so that a slow-running query doesn't prevent the JVM from shutting down

bvsharp (approved):

src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
```diff
@@ -106,4 +106,31 @@
     <sqlFile path="sql/create-mat-view-languages.sql" relativeToChangelogFile="true"/>
   </changeSet>
 
+  <!-- These slow change sets should be last, since they depend on previously run change sets! -->
+  <!-- Note: These don't do concurrent refreshes, since that requires the views to have been refreshed before -->
```
mweaver-ebsco, line 110: I only bothered with the new materialized views (the ones introduced in Q)

kjain110 (approved):

src/main/resources/db/changelog/changes/v1.1.0/changelog-v1.1.0.xml
```diff
@@ -106,4 +106,31 @@
...
+  <changeSet id="MODFQMMGR-210-refresh-payment-status" author="mweaver@ebsco.com" runOnChange="true" context="slow">
```
mweaver-ebsco, line 111: Kind of annoying to have these all in different changesets, but it makes it so if 1 fails for some reason and needs re-run, we don't have to refresh them all. It also lets us use preconditions, which is nice

PR #197, opened 2024-03-19
MODFQMMGR-181- Incorrect alpha order

kjain110:
Sorting the drop down of entity types
Sorting the drop down of field values of a particular entity type


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -67,9 +70,17 @@ public List<EntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds) {
...
+      .map(this::sortColumnsInEntityType);
   }
 
+  private EntityType sortColumnsInEntityType(EntityType entityType) {
+    List<EntityTypeColumn> sortedColumns = entityType.getColumns().stream()
+      .sorted(Comparator.comparing(EntityTypeColumn::getName))
+      .collect(Collectors.toList());
+    entityType.setColumns(sortedColumns);
+    return entityType;
+  }
```
mweaver-ebsco, line 83: Since this modifies its parameter object, I wouldn't return anything to make it obvious that this has side effects (`void` methods _always_ have side effects, or they do nothing at all, so it's a good indicator). Alternatively, this could create a new `EntityType` and return that (which I'd actually prefer since it's easier to reason about, but it's a lot more work/code)

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -110,6 +121,7 @@ private ColumnValues getFieldValuesFromEntityTypeDefinition(Field field, String
       .getValues()
       .stream()
       .filter(valueWithLabel -> valueWithLabel.getLabel().contains(searchText))
+      .sorted(Comparator.comparing(ValueWithLabel::getLabel, String.CASE_INSENSITIVE_ORDER))
```
mweaver-ebsco, line 124: Woah, I've never seen this case-insensitive order thing! Neat!

src/main/java/org/folio/fqm/service/EntityTypeService.java
ncovercash: If we use Lombok `@With` on the entity, you can just use `.withColumns` instead

ncovercash (approved):

mweaver-ebsco (approved):

PR #198, opened 2024-03-21
MODFQMMGR-207: null/empty operator doesn't work with several fields

bvsharp:
## Purpose
[MODFQMMGR-207: null/empty operator doesn't work with several fields](https://folio-org.atlassian.net/browse/MODFQMMGR-207)


ncovercash (approved):

kjain110 (approved):
LGTM

PR #210, opened 2024-03-28
MODFQMMGR-217 Fix read/write split bug

mweaver-ebsco:
This commit fixes a bug where Spring wasn't properly configured to ever use the read replica. It also fixes some minor sonar warnings related to Spring dependency injection

### DIFF
--- a/src/main/java/org/folio/fqm/repository/QueryResultsRepository.java
+++ b/src/main/java/org/folio/fqm/repository/QueryResultsRepository.java
@@ -8,6 +8,7 @@ import org.jooq.Record;
 import org.jooq.Record1;
 import org.jooq.Record2;
 import org.jooq.Table;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/fqm/repository/QueryResultsRepository.java
```diff
@@ -25,7 +26,7 @@ public class QueryResultsRepository {
   private static final Table<Record> QUERY_RESULTS_TABLE = table("query_results");
   private static final Field<UUID> QUERY_ID_FIELD = field("query_id", UUID.class);
 
-  private final DSLContext jooqContext;
+  @Qualifier("readerJooqContext") private final DSLContext jooqContext;
```
bvsharp, line 29: wouldn't this need to use the writer since it's saving results?

bvsharp (approved):

PR #218, opened 2024-03-29
MODFQMMGR-202: Include system exchange rate in Lists app

bvsharp:
## Purpose
[MODFQMMGR-202: Include system exchange rate in Lists app](https://folio-org.atlassian.net/browse/MODFQMMGR-202)


src/main/java/org/folio/fqm/repository/DataRefreshRepository.java
```diff
@@ -0,0 +1,162 @@
...
+    "INR",
+    "KRW",
+    "MXN",
+    "MYR",
+    "NZD",
+    "PHP",
+    "SGD",
+    "THB",
+    "ZAR"
+  );
```
bvsharp, line 81: These pretty much have to be hardcoded, unless we want to duplicate all the functionality in mod-finance.

src/main/java/org/folio/fqm/repository/DataRefreshRepository.java
```diff
@@ -0,0 +1,162 @@
...
+      "from", fromCurrency,
+      "to", toCurrency
+    );
+    try {
+      String exchangeRateResponse = simpleHttpClient.get(GET_EXCHANGE_RATE_PATH, exchangeRateParams);
+      DocumentContext exchangeRateInfo = JsonPath.parse(exchangeRateResponse);
+      var exchangeRate = exchangeRateInfo.read("exchangeRate");
+      if (exchangeRate instanceof BigDecimal bd) {
+        return bd.doubleValue();
+      }
```
bvsharp, line 155: the /exchange-rates API can return a double or a bigdecimal, so we have to check for that here.

mweaver-ebsco (approved):

ncovercash (approved):

PR #219, opened 2024-03-29
[MODFQMMGR-88] Initial draft at translation guidelines

ncovercash:


translations/README.md
```diff
@@ -0,0 +1,19 @@
+## Translation guidelines
+
+### General guidelines
+
+When possible, please use the following guidelines when translating field names:
+
+- Entity type names should **not** be included in the label, when it is redundant
+  - _Example_: `First name` instead of `User first name`
```
ncovercash, line 8: This is the biggest change from our current naming scheme.  However, I think it'll reduce verbosity, especially with the PoC (as we'll already be adding additional controls to handle entity type/column choosing)

(For additional context w.r.t. the PoC, I'm thinking the query builder might get an extra column like so, for a joined user + loan:)
| |  Record type | Field | Operator | Value |
| ---- | ---- | ---- | ---- | ---- |
| | User | Patron group | == | undergraduate |
| AND | Loan | Status | == | open |

translations/README.md
```diff
@@ -0,0 +1,19 @@
...
+  - _Example_: `First name`, `Effective call number`
+- Acronyms such as POL, ID, HRID, etc., should always be fully capitalized
+- As long as it is not ambiguous, use 'ID' instead of 'UUID'
+- Do not refer to arrays as 'lists' directly; instead, use plurals
+  - _Example_: `Statistical codes` instead of `Statistical code list`, `Country IDs` instead of `Country ID list`
+- For fields that build on another field's ID, disambiguate each with 'ID' and a unique description
+  - _Example_: `Material type name` and `Material type ID`
+  - **Incorrect**: `Material type` and `Material type ID`
+- The same applies for lists
+  - _Example_: `Acquisition unit IDs` and `Acquisition unit names`
```
ncovercash, line 24: These changes seem sensible to me, and not much different from our current use

translations/README.md
```diff
@@ -0,0 +1,19 @@
...
+### General guidelines
+
+When possible, please use the following guidelines when translating field names:
+
+- Entity type names should **not** be included in the label, when it is redundant
+  - _Example_: `First name` instead of `User first name`
+- Field names should have only their first word capitalized, per FOLIO UI guidelines
+  - _Example_: `First name`, `Effective call number`
+- Acronyms such as POL, ID, HRID, etc., should always be fully capitalized
+- As long as it is not ambiguous, use 'ID' instead of 'UUID'
```
ncovercash, line 12: The only place I'm unsure of this is "Instance UUID" — it seems the inventory app uses this explicitly, likely to clarify against HRID

translations/README.md
```diff
@@ -0,0 +1,24 @@
...
+  - _Example_: `First name` instead of `User first name`
+- Field names should have only their first word capitalized, per FOLIO UI guidelines
+  - _Example_: `First name`, `Effective call number`
+- Identifier acronyms such as ID, HRID, etc., should always be fully capitalized
+- As long as it is not ambiguous, use 'ID' instead of 'UUID'
+- Avoid using abbreviations/acronyms, unless they are used universally within their application
+  - _Example_: `HRID` in lieu of `Human readable identifier` because `HRID` is used universally within the inventory applications, and `Human readable identifier` is never used
+  - _Example_: `ISBN` in lieu of `International standard book number`, for the same reason
+  - _Example_: **avoid** using `Acq. unit` in lieu of `Acquisition unit`
+  - _Example_: **avoid** using `POL` instead of `Purchase order line`
```
ncovercash, line 17: How do we feel about this? This also may not matter anymore, with the first point

translations/README.md
ncovercash: Outside of the PoC, this should be fine too.  If we're making a list with "User" record type selected, "First name" already makes sense in context

mweaver-ebsco: I don't think we'd do that as part of the PoC, since everything will still be running through single entity types, from the user perspective. But if/when we support letting users create their own sets of entity types to query from, I like this idea a lot :)

translations/README.md
mweaver-ebsco: Didn't Magda say that the FOLIO convention is UUID instead of ID?

translations/README.md
bvsharp: this sounds good to me

translations/README.md
bvsharp: I think this makes sense. It's better than having some things be abbreviated and some not. The only thing I'm not sure about is POLs, I was thinking someone specifically requested that it be POL at one point.

translations/README.md
bvsharp: This one is tough. It bothers me that most fields are just ID but then we have "Instance UUID". Since firebird specifically requested that it be "Instance UUID", we'd either need to change them all IDs to UUID, or just let this be inconsistent.

translations/README.md
ncovercash: Yeah, that's the one I'm iffiest on — it's an "industry/common-knowledge term" I guess?

translations/README.md
kjain110: I thing it's better to make them as UUID. So that things remains consistent across the board

bvsharp (approved):

PR #223, opened 2024-04-03
making-const: making list as class constant

kjain110:


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -38,6 +38,12 @@
 public class EntityTypeService {
 
   private static final int COLUMN_VALUE_DEFAULT_PAGE_SIZE = 1000;
+  private static final List<String> excludedCurrencyCodes = List.of(
```
ncovercash, line 41: nit
```suggestion
  private static final List<String> EXCLUDED_CURRENCY_CODES = List.of(
```

ncovercash (approved):

PR #224, opened 2024-04-03
[MODFQMMGR-203] Implement new entity type storage system

ncovercash:
# [Jira MODFQMMGR-203](https://folio-org.atlassian.net/browse/MODFQMMGR-203)

Some things done here:
- Moved all entity types out of Liquibase and into standalone files
  - This was done by running Liquibase on a clean DB, harvesting the database table, and dumping to files
  - Harvested entity types were compared against int env with the current master deployed, to ensure none were skipped/etc.
  - As of Apr 3 @ 11a ET, all entity types are here in their previous state
- Removed all entity type definitions from historic Liquibase changesets. v1.0.4 and v1.2.0 were entirely these types of changesets, so they are now gone entirely.
- Standalone files are persisted on module install/upgrade. This is done by deleting all from the DB and re-inserting; I considered more complex update/diff logic, but for something that is relatively quick it seemed like overkill.
- [JSON5](https://json5.org/) is used for the standalone files, as it allows us to use linebreaks, comments, and other desirable features.  **Depending on your IDE, you may want to grab a JSON5 extension, since normal JSON syntax support will likely freak out.**
- Standalone files are able to be nested several directories deep, allowing nice separation (e.g. `entity-types/inventory/private/drv_item_status.json5`.  Filenames do not have to match entity type names, their only requirement is having a `.json5`/`.json` extension (so we can add .md/etc throughout, if we need to).
- Removed some lingering `labelAlias`es from two entity types
- Sorted objects to ensure consistency across entity types
- Fix the obnoxious integration smoke test to check for >0, rather than having to change it every time we add an entity type


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -76,13 +78,29 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
       .from(table(TABLE_NAME))
       .where(entityTypeIdCondition.and(publicEntityCondition))
       .fetch()
-      .map(
-        row -> new RawEntityTypeSummary(row.get(idField), row.get(nameField))
-      );
+      .map(row -> new RawEntityTypeSummary(row.get(idField), row.get(nameField)));
+  }
+
+  public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) throws JsonProcessingException {
```
ncovercash, line 89: This is done by deleting all from the DB and re-inserting; I considered more complex update/diff logic, but for an operation that is relatively quick and infrequent it seemed like overkill.

src/test/java/org/folio/fqm/repository/EntityTypeRepositoryTest.java
```diff
@@ -89,15 +89,15 @@ void shouldGetEntityTypeDefinitionWithCustomFields() {
         .visibleByDefault(true),
       new EntityTypeColumn()
         .name("custom_column_1")
-        .dataType(new BooleanType())
+        .dataType(new BooleanType().dataType("booleanType"))
```
ncovercash, line 92: I'm not sure why these weren't here before, but it was causing test issues

kjain110: you mean earlier everything was happening magically lol :)

kjain110 (approved):
LGTM

bvsharp (approved):
Nice job, this will make our lives so much better!

PR #234, opened 2024-04-12
MODFQMMGR-232: adding more finance entity type

kjain110:


mweaver-ebsco (approved):

src/main/resources/entity-types/finance/drv_fund_type_details.json5
```diff
@@ -0,0 +1,46 @@
...
+      dataType: {
+        dataType: 'rangedUUIDType',
+      },
+      queryable: true,
+      visibleByDefault: false,
+      isIdColumn: true,
+      valueGetter: "ftype.id"
+    },
+    {
+      name: '_version',
```
ncovercash, line 18: Should this be kept as `_version` in here? I thought we were reserving `_` prefixed names in FQM

mweaver-ebsco: I think we should drop the _, for the exact reason Noah said (`_` is reserved)

ncovercash (approved):

PR #238, opened 2024-04-17
MODFQMMGR-267 Only replace defined entity types when initializing

mweaver-ebsco:
This commit changes the entity type initialization process, so that only the entity types that are defined in JSON files are replaced. Any that are in the DB, but aren't defined in JSON files, will be left untouched. This prevents a non-deterministic test failure (in FqlToSqlConverterServiceIT) and also allows for custom entity types

### DIFF
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -86,21 +86,28 @@ public class EntityTypeRepository {
       .map(row -> new RawEntityTypeSummary(row.get(idField), row.get(nameField)));
   }
 
-  public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) throws JsonProcessingException {
+  public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
     log.info("Replacing entity type definitions with new set of {} entities", entityTypes.size());
 
-    // we use this instead of truncate since it has better rollback support
-    jooqContext.deleteFrom(table(TABLE_NAME)).execute();
+    jooqContext.transaction(transaction -> {
+      transaction.dsl()
+        .deleteFrom(table(TABLE_NAME))
+        .where(
+          field(ID_FIELD_NAME, UUID.class)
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -86,21 +86,29 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
...
-    InsertValuesStep2<Record, UUID, JSONB> insert = jooqContext
-      .insertInto(table(TABLE_NAME))
-      .columns(field(ID_FIELD_NAME, UUID.class), field(DEFINITION_FIELD_NAME, JSONB.class));
-
-    for (EntityType entityType : entityTypes) {
-      insert.values(UUID.fromString(entityType.getId()), JSONB.jsonb(objectMapper.writeValueAsString(entityType)));
-    }
-
-    insert.execute();
+    jooqContext.transaction(transaction -> {
```
mweaver-ebsco, line 93: This transaction made this diff much uglier than it seems. The only changes are the introduction of the transaction and the change to how this does its `deleteFrom()`

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -86,21 +86,29 @@ public List<RawEntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds)
       .map(row -> new RawEntityTypeSummary(row.get(idField), row.get(nameField)));
   }
 
-  public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) throws JsonProcessingException {
+  public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
     log.info("Replacing entity type definitions with new set of {} entities", entityTypes.size());
 
     // we use this instead of truncate since it has better rollback support
```
ncovercash, line 92: this comment can be removed now

ncovercash (approved):

PR #241, opened 2024-04-17
Entity type creator fixes

ncovercash:
entity type creator changes from #240, for master

### DIFF
--- a/entity-type-creator/components/EntityTypeFieldEditor.tsx
+++ b/entity-type-creator/components/EntityTypeFieldEditor.tsx
@@ -433,7 +433,7 @@ export default function EntityTypeFieldEditor({
               </Grid>
             </>
           ),
-          [field.valueGetter]
+          [field]
         )}
 
         {useMemo(
@@ -456,7 +456,7 @@ export default function EntityTypeFieldEditor({
               </Grid>
             </>
           ),
-          [field.filterValueGetter]
+          [field]
         )}
 
         {useMemo(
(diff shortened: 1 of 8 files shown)


entity-type-creator/components/EntityTypeFieldEditor.tsx
```diff
@@ -165,9 +165,7 @@ export default function EntityTypeFieldEditor({
               labelId={`${parentName}-${field.name}-data-type`}
               fullWidth
               value={field.dataType?.dataType}
-              onChange={(e) =>
-                onChange({ ...field, dataType: { ...field.dataType, dataType: e.target.value as DataTypeValue } })
-              }
+              onChange={(e) => onChange({ ...field, dataType: { dataType: e.target.value as DataTypeValue } })}
```
ncovercash, line 168: throws out the sub-properties of `dataType` on change, to prevent keeping `itemDataType` or `properties` when it no longer makes sense

entity-type-creator/components/EntityTypeFieldEditor.tsx
```diff
@@ -433,7 +431,7 @@ export default function EntityTypeFieldEditor({
               </Grid>
             </>
           ),
-          [field.valueGetter]
```
ncovercash, line 436: These were causing for changes in the code editor to reset the field on edit, as it kept an old copy of `field`.

kjain110 (approved):

bvsharp (approved):

PR #248, opened 2024-04-25
Working on users

kjain110:
## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist

### DIFF
--- a/src/main/resources/entity-types/users/drv_simple_user_details.json5
+++ b/src/main/resources/entity-types/users/drv_simple_user_details.json5
@@ -9,6 +9,16 @@
       dataType: {
         dataType: 'booleanType',
       },
+      values: [
+        {
+          value: 'true',
+          label: 'True',
+        },
+        {
+          value: 'false',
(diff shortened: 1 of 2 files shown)


src/main/resources/entity-types/users/drv_simple_user_details.json5
```diff
@@ -422,11 +466,19 @@
...
       valueFunction: 'lower(:value)',
     }
   ],
+  "sources": [
+    {
+      "type": "db",
+      "target": "src_users_users",
+      "alias": "user"
+    }
+  ],
```
bvsharp, line 481: Everything looks good, the only minor nit is that we can leave the double quotes off the keys (e.g. "type" could be type, "sources" could be sources). And the values could use single quotes if they don't contain single quotes within the string. So (sourceAlias: "user" could be sourceAlias: 'user'). Not a big deal but it would keep things consistent with the other fields.

PR #250, opened 2024-04-25
Making user work

kjain110:


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -110,6 +137,63 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
...
+      }).orElseThrow();
+
+    // Build entity type from sources and joins
+//    var source1 = entityType.getSources().get(0);
+//    buildCompositeEntityTypeDefinition(entityType);
+
+    return Optional.of(entityType);
+  }
+
+  private void buildCompositeEntityTypeDefinition(EntityType entityType) {
```
mweaver-ebsco, line 165: Since the code above is commented out, this method is unused

mweaver-ebsco, line 165: This method's name isn't very clear to me. When I see `build_____()`, I'd normally expect the method to return something. When I see that name but it's `void`, I have to go read the code to understand what it does

mweaver-ebsco, line 165: I could be wrong, but at a quick glance, it looks like this method can/should be `static`

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -110,6 +137,63 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
...
+    System.out.println("BUILD COMPOSITE DEFINITION");
```
mweaver-ebsco, line 166: These `System.out.println()` calls should be replaced with calls to the logger

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -110,6 +137,63 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
...
+    List<EntityTypeSourceJoin> entityTypeSourceJoins = new ArrayList<>();
+    List<EntityTypeSource> sources = entityType.getSources();
+    entityType.setFromClause("BUILD COMPOSITE DEFINITION");
```
mweaver-ebsco, line 169: Personal opinion: we should avoid using the `fromClause` entirely, so that we can remove it from the object model ASAP

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -110,6 +137,63 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
...
+      // TODO: below if-block checks if source is the first in the array, since there is nothing to join if it is.
+      //  But this needs to be done better
+      if (i == 0) { // TODO: think about replacing this with -- if (source.getJoin() == null)
+        fromClause.append(source.getAlias());
+      } else {
+        if (source.getType().equals("db")) {
+          System.out.println("Building db source");
+          EntityTypeSourceJoin join = source.getJoin();
+          String joinType = join.getType();
+          String joinCondition = join.getCondition();
```
mweaver-ebsco, line 183: Just stating the obvious here since this is implied by the comment above, but if `source.getJoin()` is null, we'll get an NPE here (and NPEs are _always_ bugs)

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -110,6 +137,63 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
...
+    StringBuilder fromClause = new StringBuilder();
+    for (int i = 0; i < sources.size(); i++) {
+      EntityTypeSource source = sources.get(i);
+      System.out.println("Building source: " + source.getAlias());
+      // TODO: below if-block checks if source is the first in the array, since there is nothing to join if it is.
+      //  But this needs to be done better
+      if (i == 0) { // TODO: think about replacing this with -- if (source.getJoin() == null)
+        fromClause.append(source.getAlias());
+      } else {
+        if (source.getType().equals("db")) {
```
mweaver-ebsco, line 179:
```suggestion
        if ("db".equals(source.getType())) {
```
This way, we handles nulls and avoid a potential NPE

src/main/java/org/folio/fqm/repository/IdStreamer.java
```diff
@@ -63,12 +69,12 @@ public int streamIdsInBatch(UUID queryId,
                               Consumer<IdsWithCancelCallback> idsConsumer) {
     UUID entityTypeId = queryDetailsRepository.getEntityTypeId(queryId);
     EntityType entityType = getEntityType(entityTypeId);
-    Condition condition = field(ID_FIELD_NAME).in(
+    Condition condition = field("\"source1\".id").in(
```
mweaver-ebsco, line 72: Don't forget to do something with this string before we merge this :)

src/main/java/org/folio/fqm/repository/IdStreamer.java
```diff
@@ -91,19 +97,29 @@ public List<List<String>> getSortedIds(String derivedTableName,
...
+                               int batchSize,
+                               Consumer<IdsWithCancelCallback> idsConsumer) {
+    String finalJoinClause = entityTypeFlatteningService.getJoinClause(entityType);
     Field<String[]> idValueGetter = IdColumnUtils.getResultIdValueGetter(entityType);
+    String finalWhereClause = sqlWhereClause.toString();
+    for (EntityTypeSource source : entityType.getSources()) {
+      String toReplace = ":" + source.getAlias();
+      String alias = "\"" + source.getAlias() + "\"";
+      finalWhereClause = finalWhereClause.replace(toReplace, alias);
+    }
```
mweaver-ebsco, line 111: We should do this replacement by using SQL parameters, if at all possible. Doing it manually is generally pretty dangerous and bug-prone. The code for handling `valueFunction` does it that way, so it could be a useful reference

src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -104,11 +109,28 @@ public List<Map<String, Object>> getResultSet(UUID entityTypeId, Fql fql, List<S
     }
 
     Condition condition = FqlToSqlConverterService.getSqlCondition(fql.fqlCondition(), entityType);
+    // TODO: might want to put next 5 lines in its own method in EntityTypeFlatteningService
```
mweaver-ebsco, line 112: Agree 100%, especially since this same code shows up in `IdStreamer`

src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -104,11 +109,28 @@ public List<Map<String, Object>> getResultSet(UUID entityTypeId, Fql fql, List<S
...
+    }
     var fieldsToSelect = getSqlFields(entityType, fields);
-    var sortCriteria = hasIdColumn(entityType) ? field(ID_FIELD_NAME) : DSL.noField();
+    String idFieldWithAlias = "";
+    String idColumnName = entityType
+      .getColumns()
+      .stream()
+      .filter(EntityTypeColumn::getIsIdColumn)
+      .findFirst()
+      .orElseThrow()
```
mweaver-ebsco, line 127: This should throw a specific exception, or at least a generic one with a clear error message

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+  public Optional<EntityType> getFlattenedEntityType(UUID entityTypeId, boolean doFinalRenames) {
+    EntityType originalEntityType = entityTypeRepository
+      .getEntityTypeDefinition(entityTypeId)
+      .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    EntityType flattenedEntityType = new EntityType()
+      .id(originalEntityType.getId())
+      .name(originalEntityType.getName())
+      ._private(originalEntityType.getPrivate())
+      .defaultSort(originalEntityType.getDefaultSort())
+      .columns(originalEntityType.getColumns()) // TODO: probably remove
```
mweaver-ebsco, line 41: Personal opinion: don't remove it. If we keep the columns, it makes it possible to augment an existing entity type by importing it and adding some fields

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+      .sourceView(originalEntityType.getSourceView()) // Possibly unneeded
+      .sourceViewExtractor(originalEntityType.getSourceViewExtractor()); // Possibly unneeded
+
+    List<EntityTypeColumn> finalColumns = new ArrayList<>();
+    for (EntityTypeSource source : originalEntityType.getSources()) {
+      if (source.getType().equals("db")) {
+        Pair<EntityTypeSource, List<EntityTypeColumn>> updatePair = getConvertedSourceAndColumns(originalEntityType, source, null, false); // TODO: think about this, may not be able to hardcode false here
+        flattenedEntityType.addSourcesItem(updatePair.component1());
+        finalColumns.addAll(updatePair.component2());
+      } else {
```
mweaver-ebsco, line 55: I'd probably check for "entity-type" here and throw an exception when it's not one of the valid values.

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
+package org.folio.fqm.service;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import kotlin.Pair;
```
mweaver-ebsco, line 4: We should probably use a different implementation of `Pair`. We're not actually using Kotlin (unfortunately), so it just feels weird to use stuff from its standard lib

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+      .customFieldEntityTypeId(originalEntityType.getCustomFieldEntityTypeId())
+      .labelAlias(originalEntityType.getLabelAlias())
+      .root(originalEntityType.getRoot())
+      .sourceView(originalEntityType.getSourceView()) // Possibly unneeded
+      .sourceViewExtractor(originalEntityType.getSourceViewExtractor()); // Possibly unneeded
+
+    List<EntityTypeColumn> finalColumns = new ArrayList<>();
+    for (EntityTypeSource source : originalEntityType.getSources()) {
+      if (source.getType().equals("db")) {
+        Pair<EntityTypeSource, List<EntityTypeColumn>> updatePair = getConvertedSourceAndColumns(originalEntityType, source, null, false); // TODO: think about this, may not be able to hardcode false here
```
mweaver-ebsco, line 52: Comments like this make me wish we _were_ using something where you can name parameters at the call-site, like kotlin. I have no idea what the false does here, which is something that could be fixed at the language-level if Java supported it ☹️

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+        finalColumns.addAll(updatePair.component2());
+      } else {
+        UUID sourceEntityTypeId = UUID.fromString(source.getId());
+        EntityType flattenedSourceDefinition = getFlattenedEntityType(sourceEntityTypeId, false)
+          .orElseThrow(() -> new EntityTypeNotFoundException(sourceEntityTypeId));
+
+        // If an entity type source contains multiple db sources, then we need to keep the original alias in order to
+        // distinguish the different targets. Frequently, it will likely only have one db source. In this case we
+        // can use the outer alias only, in order to keep field names more concise
+        boolean keepOriginalAlias = countDbSources(flattenedSourceDefinition) > 1;
```
mweaver-ebsco, line 63: Personal opinion, but I could probably be swayed pretty easily: don't bother with this. I'd prefer less concise field names if it means simpler code and more consistent behavior.

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+
+    flattenedEntityType.columns(finalColumns);
+    if (doFinalRenames) {
+      List<EntityTypeColumn> convertedColumns = finalColumnConversion(flattenedEntityType);
+      flattenedEntityType.columns(convertedColumns);
+    }
+    return Optional.of(flattenedEntityType);
+  }
+
+  public String getJoinClause(EntityType flattenedEntityType) {
```
mweaver-ebsco, line 86: Nit: I don't like this name, since afaik, there's no such thing as a join clause

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+
+    for (EntityTypeSource source : entityType.getSources()) {
+      if (!visited.contains(source.getAlias())) {
+        dfs(source, sourceMap, visited, orderedList);
+      }
+    }
+    return orderedList;
+  }
+
+  private static void dfs(EntityTypeSource source, Map<String, EntityTypeSource> sourceMap, Set<String> visited, List<EntityTypeSource> orderedList) {
```
mweaver-ebsco, line 146: Not a fan of this name, personally. I had to actually read the code to understand the name

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+    if (source.getJoin() != null) {
+      EntityTypeSource joinToSource = sourceMap.get(source.getJoin().getJoinTo());
+      if (!visited.contains(joinToSource.getAlias())) {
+        dfs(joinToSource, sourceMap, visited, orderedList);
+      }
+    }
+    orderedList.add(source);
+  }
+
+  private void updateOtherSources(String oldSourceName, String newSourceName, List<EntityTypeSource> otherSources) {
```
mweaver-ebsco, line 157: This method name is killing me, too 😛 I dunno what "other sources" are, so it might as well be named `doThings()` 🙂

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+    } catch (Exception e) {
+      return new EntityTypeColumn(); // TODO: do something better here
+    }
+  }
+
+  private long countDbSources(EntityType entityType) {
+    return entityType
+      .getSources()
+      .stream()
+      .filter(source -> !Boolean.TRUE.equals(source.getFlattened()) && source.getType().equals("db"))
```
mweaver-ebsco, line 239: Maybe set the type to "flattened" or something and get rid of the `flattened` property? That'd simplify the object model a bit and probably simplify some code outside this method, too, too

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -0,0 +1,242 @@
...
+  private EntityTypeColumn copyColumn(EntityTypeColumn column) {
+    try {
+      String json = objectMapper.writeValueAsString(column);
+      return objectMapper.readValue(json, EntityTypeColumn.class);
+    } catch (Exception e) {
+      return new EntityTypeColumn(); // TODO: do something better here
+    }
+  }
+
+  private long countDbSources(EntityType entityType) {
```
mweaver-ebsco, line 235:
```suggestion
  private static long countDbSources(EntityType entityType) {
```

PR #255, opened 2024-04-29
MODFQMMGR-246: Create POC complex entity types Ledger #5-6

bvsharp:
## Purpose
[MODFQMMGR-246: Create POC complex entity types Ledger #5-6](https://folio-org.atlassian.net/browse/MODFQMMGR-246)


src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -75,7 +75,8 @@ public List<Map<String, Object>> getResultSet(UUID entityTypeId,
       if (columnDataType.equals("rangedUUIDType") || columnDataType.equals("openUUIDType")) {
         List<UUID> idColumnValuesAsUUIDs = idColumnValues
           .stream()
-          .map(UUID::fromString)
+          // TODO: the below check keeps things just-about working, but still leads to records being counted as "deleted" if one of their id columns is null. Need to handle this better.
```
bvsharp, line 78: I figured I'd just leave a comment for now rather than spend any time addressing this, especially because it's not really related to the POC.

ncovercash: What if we did something goofy like map to "null" or ""?

ncovercash (approved):

mweaver-ebsco (approved):

kjain110 (approved):

PR #262, opened 2024-05-03
Loan and associates

ncovercash:


kjain110 (approved):

entity-type-creator/components/Importer/ImportStep.tsx
```diff
@@ -131,6 +150,11 @@ function getSimpleTypeOf(schema: Schema) {
           return 'date' as const;
         } else if ('format' in schema && schema.format === 'uuid') {
           return 'uuid' as const;
+        } else if (
+          'pattern' in schema &&
+          schema.pattern === '^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$'
```
mweaver-ebsco, line 155: Nit: At first glance, this regex is kind of gnarly. It might be worthwhile to put it in a const somewhere, just to give it a name. Either that or add a short comment to save someone the 5 seconds it takes to realize it matches UUIDs :P

ncovercash: it's awful, but it's used a lot in RAMLs (alongside other ones that `$ref: ../whatever/uuid.json`???)

ncovercash: also idk i feel like making it a constant elsewhere would just make it more confusing — it's only ever used in that one spot

mweaver-ebsco (approved):

PR #265, opened 2024-05-08
MODFQMMGR-247 & MODFQMMGR-248: Add complex entity types

bvsharp:
## Purpose
[MODFQMMGR-247](https://folio-org.atlassian.net/browse/MODFQMMGR-247) & [MODFQMMGR-248](https://folio-org.atlassian.net/browse/MODFQMMGR-248): Add complex entity types and GROUP BY logic needed to support them


src/main/resources/entity-types/invoice/composite_invoice__voucher_line__fund_distribution.json5
```diff
@@ -0,0 +1,59 @@
...
+  id: '087bbe42-cc6d-40a9-8f41-c5fd6cee6037',
+  name: 'composite_invoice__voucher_line__fund_distribution',
+  private: false,
+  sources: [
+    {
+      id: 'b4c76df5-d730-4ede-8a98-ab04c5e222ab',
+      type: 'entity-type',
+      alias: 'voucher_line__fund_distribution',
+      useIdColumns: true,
+    },
```
bvsharp, line 11: I know we talked about keeping the entity types as flat as possible, but I thought it might be good to have one composite entity type containing another composite one, just to show that it's possible.

src/main/resources/entity-types/users/drv_simple_user_details.json5
```diff
@@ -450,7 +450,7 @@
       )",
     },
     {
-      name: 'departments',
```
kjain110, line 453: can we add translation for ids and name in en.json file (for simple user one)

bvsharp:
> No translation changes?
Since the translations aren't working at all for complex ones, I figured that could be handled in the scope of the translation ticket. That way we can at least get these in and tested during this sprint.

kjain110 (approved):

PR #271, opened 2024-05-15
MODFQMMGR-274: Add translations for composite entity types

bvsharp:
## Purpose
Add support for translating composite entity types. This is a very rough implementation and does add some tech debt for when we merge the poc branch into master, but it should be good enough for now, given that we have other priorities.


kjain110 (approved):
LGTM.

ncovercash (approved):

PR #285, opened 2024-06-11
MODFQMMGR-308:Keep UUID fields as UUIDs to take advantage of indexes

kjain110:
## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -32,7 +30,7 @@
 
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
-import java.util.List;
+import java.util.*;
```
ncovercash, line 33: :(

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -18,11 +18,9 @@
 import org.folio.fql.service.FqlService;
 import org.folio.fql.service.FqlValidationService;
 import org.folio.fqm.exception.FieldNotFoundException;
+import org.folio.fqm.exception.InvalidFqlException;
 import org.folio.fqm.utils.SqlFieldIdentificationUtils;
-import org.folio.querytool.domain.dto.DateType;
-import org.folio.querytool.domain.dto.EntityDataType;
-import org.folio.querytool.domain.dto.EntityType;
-import org.folio.querytool.domain.dto.Field;
+import org.folio.querytool.domain.dto.*;
```
ncovercash, line 23: :(

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -135,7 +141,17 @@ private static Condition handleEquals(EqualsCondition equalsCondition, EntityTyp
...
+    if (RANGED_UUID_TYPE.equals(filterFieldDataType) || OPEN_UUID_TYPE.equals(filterFieldDataType)) {
+      String value = (String) equalsCondition.value();
+      if (!isValidUUID(value)) {
+        Map<String, String> errors = Map.of(INVALID_UUID, ERROR_MESSAGE + value);
+        throw new InvalidFqlException(field + equalsCondition.toString(), errors);
+      }
+      return cast(field, UUID.class).eq(cast(value, UUID.class));
+    }
+
+    if (STRING_TYPE.equals(dataType) || RANGED_UUID_TYPE.equals(dataType) || OPEN_UUID_TYPE.equals(dataType) || DATE_TYPE.equals(dataType) || STRING_UUID_TYPE.equals(dataType))  {
```
bvsharp, line 154: We don't need the RANGED_UUID_TYPE or OPEN_UUID_TYPE conditions here anymore, do we? Since they'll be handled above.

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -146,7 +162,16 @@ private static Condition handleNotEquals(NotEqualsCondition notEqualsCondition,
...
+    String filterFieldDataType = getFieldForFiltering(notEqualsCondition, entityType).getDataType().getDataType();
+    if (RANGED_UUID_TYPE.equals(filterFieldDataType) || OPEN_UUID_TYPE.equals(filterFieldDataType) ) {
+      String value = (String) notEqualsCondition.value();
+      if (!isValidUUID(value)) {
+        Map<String, String> errors = Map.of(INVALID_UUID, ERROR_MESSAGE + value);
+        throw new InvalidFqlException(field + notEqualsCondition.toString(), errors);
+      }
+      return cast(field, UUID.class).eq(cast(value, UUID.class));
+    }
+    if (STRING_TYPE.equals(dataType) || RANGED_UUID_TYPE.equals(dataType) || OPEN_UUID_TYPE.equals(dataType) || DATE_TYPE.equals(dataType) || STRING_UUID_TYPE.equals(dataType)) {
```
bvsharp, line 174: We don't need the RANGED_UUID_TYPE or OPEN_UUID_TYPE conditions here anymore, do we? Since they'll be handled above.

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -358,4 +402,8 @@ private static <T> org.jooq.Field<T> valueField(T value, FieldCondition<?> condi
     }
     return val(value);
   }
+  public static boolean isValidUUID(String uuidStr){
+    UUID uuid = UUID.fromString(uuidStr);
+    return uuid.toString().equals(uuidStr);
+  }
```
bvsharp, line 408: Does this work for checking whether something is a UUID? Seems like it might just throw an IllegalArgumentException if it doesn't match.

bvsharp: You could make the method void and do a try-catch block that catches IllegalArgumentException, and throws InvalidFqlException from the the catch block. Then you could call that method from the outer methods.

ncovercash (approved):

bvsharp (approved):

PR #286, opened 2024-06-12
MODFQMMGR-308-chaning-datatype: changing dt to stringUUIDType

kjain110:


ncovercash (approved):

src/main/resources/entity-types/organizations/drv_organization_details.json5
```diff
@@ -282,7 +282,7 @@
     {
       name: 'tax_id',
       dataType: {
-        dataType: 'stringType',
+        dataType: 'stringUUIDType',
```
bvsharp, line 285: This field is actually just a string so it can stay stringType.

src/main/resources/entity-types/users/drv_user_details.json5
```diff
@@ -227,7 +227,7 @@
     {
       name: 'user_external_system_id',
       dataType: {
-        dataType: 'stringType',
+        dataType: 'stringUUIDType',
```
bvsharp, line 230: Same as above, this field is actually just a string so it can stay stringType.

mweaver-ebsco (approved):

PR #306, opened 2024-06-28
MODFQMMGR-371 Add support for complex data types in composite ETs

mweaver-ebsco:


ncovercash (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -221,21 +224,61 @@ private Pair<EntityTypeSource, List<EntityTypeColumn>> getConvertedSourceAndColu
...
-      valueGetter = valueGetter.replace(toReplace, sourceAlias);
-      if (filterValueGetter != null) {
-        filterValueGetter = filterValueGetter.replace(toReplace, sourceAlias);
-      }
-      column.valueGetter(valueGetter);
-      column.filterValueGetter(filterValueGetter);
-      finalColumns.add(column);
+    return flattenedEntityType.getColumns()
+      .stream()
+      .map(column -> injectSourceAlias(column, column.getSourceAlias()))
```
mweaver-ebsco, line 229: The sourceAlias gets pulled out here, instead of in `injectSourceAlias()` because that method takes a `T extends Field` parameter, and the sourceAlias comes from `EntityTypeColumn` objects, which is a subclass of `Field`. By using `T extends Field`, it lets `injectSourceAliasForObjectType()` recursively call `injectSourceAlias` with the properties from the object type

bvsharp (approved):

kjain110 (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
ncovercash: Any reason not to just use Field?

PR #307, opened 2024-06-28
MODFQMMGR-369 Add "Unknown" as a selectable value for preferred contact type

mweaver-ebsco:


src/main/resources/entity-types/users/simple_user_details.json5
```diff
@@ -158,6 +158,10 @@
           value: 'Text Message',
           label: 'Text Message',
         },
+        {
+          value: 'unknown',
+          label: 'Unknown',
+        },
```
mweaver-ebsco, line 164: We replace null values with "unknown" for this field, but weren't providing a way for users to actually query for that (is-empty doesn't work because of our substitution, and the hard-coded list of values here makes it so there's no way to users to search for something outside the list)

ncovercash (approved):

bvsharp (approved):

kjain110 (approved):

PR #316, opened 2024-07-10
[MODFQMMGR-387] Pre-resolve locale-related information before starting async tasks

ncovercash:
# [MODFQMMGR-387](https://folio-org.atlassian.net/browse/MODFQMMGR-387/)

The localization service uses information from the current thread to determine which locale(s) should be used; this normally works fine, however, Spring discards this information when we are processing queries asynchronously, causing a large quantity of logged errors. This does not affect the results (since the translated names are not used for query processing), however, it is annoying to look at.

It [is possible to share context](https://stackoverflow.com/a/50138897/4236490) with async threads in Spring, however, just doing this alone causes weirder issues where Tomcat will discard the request context once the request completes, causing it to no longer be accessible when the async thread attempts to access it — I was unable to find any way around this without disabling request recycling entirely, which is asking for memory leaks (and [Spring maintainers specifically mention that this context will be unreliable outside the thread](https://github.com/spring-projects/spring-boot/issues/36763#issuecomment-1667382305)).

To solve this, I'm just passing the fully resolved entity type as part of the query task, so no resolution needs to be done inside the async task. This is already being resolved/flattened/translated before the job is kicked off, it was just redone in the task, so, if anything, this should be a performance increase.

### DIFF
--- a/src/test/java/org/folio/fqm/utils/IdStreamerTest.java
+++ b/src/test/java/org/folio/fqm/utils/IdStreamerTest.java
@@ -2,7 +2,6 @@ package org.folio.fqm.utils;
 
 import static org.folio.fqm.utils.IdStreamerTestDataProvider.TEST_CONTENT_IDS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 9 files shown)


src/test/java/org/folio/fqm/utils/IdStreamerTest.java
```diff
@@ -109,34 +103,4 @@ void shouldGetSortedIds() {
     );
     assertEquals(expectedIds, actualIds);
   }
-
-  // TODO: this test is kind of butchered. Clean up if possible
-  @Test
-  void shouldThrowExceptionWhenEntityTypeNotFound() {
```
ncovercash, line 115: IdStreamer no longer fetches the ET itself => no need for this test

mweaver-ebsco (approved):

bvsharp (approved):

PR #317, opened 2024-07-11
MODFQMMGR-368-fixes: changing the dash separator

kjain110:


src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -69,7 +69,7 @@ private String getSourceTranslationPrefix(EntityType entityType, String columnNa
     if (currentSourceIndex > 0) {
       String currentSource = columnName.substring(0, currentSourceIndex);
       String formattedKey = ENTITY_TYPE_COLUMN_AND_SOURCE_LABEL_TRANSLATION_TEMPLATE.formatted(entityType.getName(), currentSource);
-      return  translationService.format(formattedKey) + " - ";
+      return  translationService.format(formattedKey) + " — ";
```
kjain110, line 72: is the dash used here is correct ?

mweaver-ebsco: If I'm reading MODFQMMGR-368 correctly, the separator needs to be localized and the `en` version should be the em dash

ncovercash: monospace is hard:
      return  translationService.format(formattedKey) + " - ";
      return  translationService.format(formattedKey) + " — ";

it's good

ncovercash (approved):

src/main/java/org/folio/fqm/service/LocalizationService.java
ncovercash: this whole thing (dash/combining) should be changed at some point though, to accommodate other languages and RTL, but that's later

PR #328, opened 2024-07-16
Modfqmmgr 391- Add endpoint to get latest query version

kjain110:


src/main/resources/swagger.api/mod-fqm-manager.yaml
```diff
@@ -62,7 +62,23 @@ paths:
           $ref: '#/components/responses/badRequestResponse'
         '500':
           $ref: '#/components/responses/internalServerErrorResponse'
-
+  /query/migrate/version:
+    get:
+      operationId: getQueryVersion
+      tags:
+        - queryVersion
+      description: Get the version of the query.
```
mweaver-ebsco, line 70: What query?

src/main/java/org/folio/fqm/service/MigrationService.java
```diff
@@ -1,8 +1,13 @@
...
+
+import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.NotImplementedException;
+import org.springframework.stereotype.Service;
 
+@Service
+@RequiredArgsConstructor
 public class MigrationService {
 
   public int getLatestVersion() {
```
mweaver-ebsco, line 13: Out of scope for this PR, probably, but I still think we should use something besides `int` for the version. I don't really want to have to remember to bump another version somewhere when we're already having to juggle the artifact version, API versions, and liquibase version

kjain110: something like this ```
@Value("${artifact.version}")
    private String artifactVersion;
```

File: src/main/resources/swagger.api/mod-fqm-manager.yaml [hunk 3]
```diff
@@ -62,7 +62,23 @@ paths:
...
-
+  /fqm/version:
+    get:
+      operationId: getQueryVersion
+      tags:
+        - queryVersion
+      description: Get the version of the fqm query.
+      responses:
+        '200':
+          description: 'Version of the fqm query'
```
mweaver-ebsco, line 73: Small thing, but I'd update these to no longer refer to query, since the version covers more than just queries (queries are the things that get updated, based on the FQM version, so the version is associated with FQM and not the query)

mweaver-ebsco (approved):

bvsharp (approved):

PR #329, opened 2024-07-17
MODFQMMGR-339: Include ECS columns in query builder

bvsharp:
## Purpose
[MODFQMMGR-339: Extend entity type response to add additional data column for query builder](https://folio-org.atlassian.net/browse/MODFQMMGR-339)

This story makes the new ECS columns available for use from the query builder.

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
@@ -8,8 +8,14 @@ import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Stream;
+
+import com.jayway.jsonpath.DocumentContext;
+import com.jayway.jsonpath.JsonPath;
+import feign.FeignException;
 import lombok.extern.log4j.Log4j2;
+import org.folio.fqm.client.SimpleHttpClient;
 import org.folio.fqm.repository.EntityTypeRepository;
 import org.folio.querytool.domain.dto.EntityType;
 import org.folio.spring.FolioExecutionContext;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
```diff
@@ -62,7 +70,16 @@ public EntityTypeInitializationService(
...
+    String centralTenantId = "${central_tenant_id}";
+    try {
+      String rawJson = ecsClient.get("consortia-configuration", Map.of("limit", String.valueOf(100)));
+      DocumentContext parsedJson = JsonPath.parse(rawJson);
+      centralTenantId = parsedJson.read("centralTenantId");
+      log.info("ECS central tenant ID: {}", centralTenantId);
+    } catch (Exception e) {
+      log.info("ECS is not enabled for tenant {}", folioExecutionContext.getTenantId());
+    }
+    String finalCentralTenantId = centralTenantId;
```
bvsharp, line 82: Necessary to get around "Variable used in lambda expression should be final or effectively final"

kjain110 (approved):

src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
```diff
@@ -62,7 +70,16 @@ public EntityTypeInitializationService(
...
   public void initializeEntityTypes() throws IOException {
     log.info("Initializing entity types");
-
+    String centralTenantId = "${central_tenant_id}";
+    try {
+      String rawJson = ecsClient.get("consortia-configuration", Map.of("limit", String.valueOf(100)));
+      DocumentContext parsedJson = JsonPath.parse(rawJson);
+      centralTenantId = parsedJson.read("centralTenantId");
+      log.info("ECS central tenant ID: {}", centralTenantId);
+    } catch (Exception e) {
```
mweaver-ebsco, line 79: Is there a better exception we can catch here? I worry about this potentially hiding bugs if something unexpected happens

src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
```diff
@@ -74,7 +91,8 @@ public void initializeEntityTypes() throws IOException {
           return objectMapper.readValue(
             resource
               .getContentAsString(StandardCharsets.UTF_8)
-              .replace("${tenant_id}", folioExecutionContext.getTenantId()),
+              .replace("${tenant_id}", folioExecutionContext.getTenantId())
+              .replace("${central_tenant_id}", finalCentralTenantId),
```
mweaver-ebsco, line 95: Do we need to do this in the tenant management service, too? That'd get us consistent behavior between liquibase and entity type definitions, but I'm not sure if that's necessary or not

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/EntityTypeInitializationService.java
bvsharp: I didn't think that was necessary at the moment since there's nowhere within the liquibase scripts where we actually needed the central tenant id, and I can't think of a situation where we would need it. It shouldn't be a big deal to add later if we need to though.

ncovercash: good catch...the potential dependency concerns scare me lol

PR #333, opened 2024-07-18
MODFQMMGR-394-Retrieve JSONB (entire records) in Member Tenant

kjain110:


mweaver-ebsco (approved):

translations/mod-fqm-manager/en.json
```diff
@@ -200,6 +200,7 @@
   "entityType.drv_fund_type_details.version": "Version",
   "entityType.simple_group_details": "Group",
   "entityType.simple_group_details.created_by_user_id": "Created by user UUID",
+  "entityType.simple_group_details.groups_jsonb": "JSONB",
```
ncovercash, line 203: do we need translations if these are never exposed to the user?

ncovercash (approved):

translations/mod-fqm-manager/en.json
kjain110: right now there is no way to hide it from user until we implement that hidden field story

src/main/resources/entity-types/circulation/simple_loan.json5
```diff
@@ -368,5 +368,16 @@
       visibleByDefault: false,
       valueGetter: ":sourceAlias.jsonb->'reminders'->'lastFeeBilled'->>'date'",
     },
+    {
+      name: 'loan_jsonb',
```
bvsharp, line 372: Is there a reason we can't just call each column 'jsonb' (instead of prepending it with the entity type name)? None of the other columns use the prepending, and the translations aren't using it either.

mweaver-ebsco (approved):

bvsharp (approved):

PR #336, opened 2024-07-23
[MODFQMMGR-389] Migration service implementation

ncovercash:
# [MODFQMMGR-389](https://folio-org.atlassian.net/browse/MODFQMMGR-389)

## Purpose
Develop a robust, future-proof system to protect queries, entity type IDs, and field lists against the forces of time.

## Approach
Implementation of `MigrationService` and set of methods which take abstract `MigrationStrategy`s, enabling simple migrations with little friction, or extremely complex ones, to both be completed.

Additionally, a tool was added to the entity type creator to aid in mapping columns between entity types for simple migrations.

### Migration strategies

For simple entity type/column changes, we provide a `AbstractSimpleMigrationStrategy` which can be easily extended:

```java
public class MyMigration extends AbstractSimpleMigrationStrategy {
  @Override public String getLabel() { return "My super cool breaking change"; }
  @Override public String getSourceVersion() { return "341"; }
  @Override public String getTargetVersion() { return "342"; }

  @Override
  protected Map<UUID, UUID> getEntityTypeChanges() {
    return Map.of(OLD ENTITY TYPE ID, NEW ENTITY TYPE ID);
  }

  @Override
  protected Map<UUID, UUID> getEntityTypeChanges() {
    return Map.of(OLD ENTITY TYPE ID, Map.of("id", "id_but_cooler"),
                            SOMETHING ELSE, Map.of("name", "name but i changed it too for some reason"));
  }
}
```

Or, for more complex changes, `MigrationStrategy` can be extended directly:

```java
public class BigBreaker implements MigrationStrategy {
  @Override public String getLabel() { return "$eq is no fun"; }
  
  @Override 
  public boolean applies(FqlService, MigratableQueryInformation src) {
    return src.fqlQuery().contains("$eq");
    // FqlService is provided here to facilitate proper deserialization. The other way I could think of was to make all of these spring components, but some part of me died with this solution, so.
    // this would be smarter to be something like this, since we should use our version numbers, but.
    return {magic deserialization}.getVersion().equals("412");
  }

  @Override
  public MigratableQueryInformation apply(FqlService, MigratableQueryInformation src) {
    return src.withFqlQuery(src.fqlQuery().replaceAll("$eq", "$neq"));
  }
}
```

One important note here is that **after `apply`ing a migration to a query, that migration should no longer be eligible (`applies` returns false)**. The smart way to do this is by changing version numbers, but this is left open-ended, should someone want to footgun.

### Migration helper tool

This is a tool which allows a user to quickly draw connections between old/new entity types and get copy-pastable migration code. As the entity type creator has not been updated to resolve sources, this requires pasting in the new and old entity types, with full flattening:

https://github.com/user-attachments/assets/68a36d14-e872-4e6b-a8d5-74f6efc05775

## V0->V1 migration (PoC)

These were all based on the [final changes merged](https://github.com/folio-org/mod-fqm-manager/commit/9a1e180186dc7178b0c3416f1aec088c9ce39a54).  Here is a summary of what changed, and the questions I have:

| Migration Status   | Source File                                                   | New Entity Type                     | Notes                                                                                                                                                                           |
| ------------------ | ------------------------------------------------------------- | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| :warning:          | `circulation/drv_loan_details.json5`                          | `composite_loan_details`            | Unsure how to migrate the primary contributor, as the new ET has an array of contributors, rather than just one                                                                 |
| :x:                | `circulation/private/drv_loan_status.json5`                   | _NONE_                              | Original materialized view is now unused (noted in [MODFQMMGR-404](https://folio-org.atlassian.net/browse/MODFQMMGR-404))                                                       |
| :white_check_mark: | `circulation/private/src_circulation_loan_policy.json5`       | `simple_loan_policy`                |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/drv_holdings_record_details.json5`                 | `composite_holdings_record`         | Same ID as original ET                                                                                                                                                          |
| :white_check_mark: | `inventory/drv_instances.json5`                               | `composite_instances`               | Same ID as original ET; individual contributor arrays mapped to new object structure                                                                                            |
| :warning:          | `inventory/drv_item_details.json5`                            | `composite_item_details`            | Unsure how to migrate the primary contributor, as the new ET has an array of contributors, rather than just one                                                                 |
| N/A                | `inventory/private/drv_inventory_statistical_code_full.json5` | `simple_statistical_code_full`      | No ID or column changes                                                                                                                                                         |
| :x:                | `inventory/private/drv_item_callnumber_location.json5`        | _NONE_                              |                                                                                                                                                                                 |
| :x::warning:       | `inventory/private/drv_item_holdingsrecord_instance.json5`    | _NONE_                              | This seems like a weird one to lose? Is it replaced with something else?                                                                                                        |
| :x:                | `inventory/private/drv_item_status.json5`                     | _NONE_                              | Original materialized view is now unused (noted in [MODFQMMGR-404](https://folio-org.atlassian.net/browse/MODFQMMGR-404))                                                       |
| :white_check_mark: | `inventory/private/src_inventory_call_number_type.json5`      | `simple_call_number_type`           |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_contributor_name_type.json5` | `drv_contributor_name_type_details` |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_contributor_type.json5`      | `drv_contributor_type_details`      |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_instance_status.json5`       | `simple_instance_status`            |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_location.json5`              | `simple_locations`                  |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_loclibrary.json5`            | `simple_loclibrary`                 |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_material_type.json5`         | `simple_material_type_details`      |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_mode_of_issuance.json5`      | `simple_mode_of_issuance`           |                                                                                                                                                                                 |
| :white_check_mark: | `inventory/private/src_inventory_service_point.json5`         | `simple_service_point_details`      |                                                                                                                                                                                 |
| :white_check_mark: | `organizations/drv_organization_contacts.json5`               | `simple_organization`               |                                                                                                                                                                                 |
| :white_check_mark: | `organizations/drv_organization_details.json5`                | `simple_organization`               |                                                                                                                                                                                 |
| :white_check_mark: | `organizations/private/src_organization_types.json5`          | `simple_organization_types`         |                                                                                                                                                                                 |
| :white_check_mark: | `organizations/private/src_organizations.json5`               | `simple_organization`               |                                                                                                                                                                                 |
| :x:                | `pol/drv_pol_payment_status.json5`                            | _NONE_                              | Original materialized view is now unused (noted in [MODFQMMGR-404](https://folio-org.atlassian.net/browse/MODFQMMGR-404))                                                       |
| :x:                | `pol/drv_pol_receipt_status.json5`                            | _NONE_                              | Original materialized view is now unused (noted in [MODFQMMGR-404](https://folio-org.atlassian.net/browse/MODFQMMGR-404))                                                       |
| :white_check_mark: | `pol/drv_purchase_order_line_details.json5`                   | `composite_purchase_order_lines`    |                                                                                                                                                                                 |
| N/A                | `shared/{private/src_acquisitions_unit.json5`                 | `simple_acquisitions_unit`          | No ID or column changes                                                                                                                                                         |
| N/A                | `shared/{private/drv_languages.json5`                         | `simple_languages`                  | No ID or column changes                                                                                                                                                         |
| :warning:          | `users/drv_user_details.json5`                                | `composite_user_details`            | Unsure how to migrate the primary address, as the new ET has an array of addresses, rather than just one. Individual address field array columns were mapped to new object ones |
| :white_check_mark: | `users/private/src_users_addresstype.json5`                   | `simple_address_types`              |                                                                                                                                                                                 |
| :white_check_mark: | `users/private/src_users_departments.json5`                   | `simple_department_details`         |                                                                                                                                                                                 |
| :white_check_mark: | `users/private/src_users_groups.json5`                        | `simple_group_details`              |                                                                                                                                                                                 |

Please note that these in several cases **did not match** the table at https://eis.atlassian.net/wiki/spaces/PRO/pages/751337911/POC+to+Ramsons+Release+Key+Considerations; I'm unsure if that's due to outdated information (and several rows there seemed to reflect mid-PoC changes, which are not included here).

#### TODOS and Open Questions
- [ ] Several in the PoC migration (see table above and/or code)
- [ ] I don't like having to enumerate each strategy in `MigrationStrategyRepository` (that's the only reason that got broken into its own file), but the only alternative I can see is some annotation sneakiness, or using reflection, and ick.


src/main/java/org/folio/fqm/service/MigrationService.java
```diff
@@ -1,28 +1,50 @@
...
 
   public MigratableQueryInformation migrate(MigratableQueryInformation migratableQueryInformation) {
-    // return migratableQueryInformation;
-    throw new NotImplementedException();
-  }
+    while (isMigrationNeeded(migratableQueryInformation)) {
+      for (MigrationStrategy strategy : migrationStrategyRepository.getMigrationStrategies()) {
+        if (strategy.applies(migratableQueryInformation)) {
+          log.info("Applying {} to {}", strategy.getLabel(), migratableQueryInformation);
+          migratableQueryInformation = strategy.apply(migratableQueryInformation);
```
ncovercash, line 43: Inefficient. Some, particularly algorithms professors, might even say `O(n^2)`.

However, this is not run frequently. Additionally, most of these `.applies` calls should be a simple `.equals`. Something cleverer, such as a `Map` from source versions to `MigrationStrategy` could work, however, there may be cases where we want to perform a migration on a range of versions, or something more strange, so the extensibility is left here.

Additionally, if `migrationStrategyRepository` supplies migrators in the order from V0 -> V∞, this loop should theoretically only need to execute once.

src/main/java/org/folio/fqm/migration/strategies/V0POCMigration.java
mweaver-ebsco, line 1: I wonder if it'd be worthwhile to create a DB-backed migration strategy, where it reads all of these migration details from the DB instead of having it hard-coded in the app code. I like the idea of keeping the infrastructure to let us define complex migrations in code, but for simple stuff like this strategy, it feels more like data/config. Not really a big deal, by any means, but it might be worth considering eventually

src/main/java/org/folio/fqm/migration/strategies/V0POCMigration.java
```diff
@@ -0,0 +1,84 @@
...
+  protected Map<UUID, UUID> getEntityTypeChanges() {
+    return Map.ofEntries(
+      Map.entry(OLD_DRV_LOAN_DETAILS, NEW_COMPOSITE_LOAN_DETAILS),
+      Map.entry(OLD_SRC_CIRCULATION_LOAN_POLICY, NEW_SIMPLE_LOAN_POLICY)
+    );
+  }
+
+  @Override
+  protected Map<UUID, Map<String, String>> getFieldChanges() {
+    return Map.ofEntries(
```
mweaver-ebsco, line 44: Small thing, but this map looks kind of expensive to build, since it's has to instantiate a ton of objects every time (especially considering it doesn't look like the full set of fields, so the final version will be MUCH bigger). It may be worth putting into a `static final` class  field, to make this cheaper

src/main/java/org/folio/fqm/service/MigrationService.java
```diff
@@ -1,28 +1,50 @@
...
-    // return 1;
-    throw new NotImplementedException();
+  private static final String CURRENT_VERSION = "1";
+
+  private final FqlService fqlService;
+  private final MigrationStrategyRepository migrationStrategyRepository;
+
+  public String getLatestVersion() {
+    return CURRENT_VERSION;
   }
```
mweaver-ebsco, line 24: Just throwing this out there 🙂 

```java
  private final BuildProperties buildProperties; // Hand wavy Spring Boot magic
  public String getLatestVersion() {
    return buildProperties.getVersion();
  }
```

src/main/java/org/folio/fqm/migration/MigrationStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.function.BiFunction;
+import org.folio.fql.service.FqlService;
+
+public interface MigrationStrategy
+  extends BiFunction<FqlService, MigratableQueryInformation, MigratableQueryInformation> {
+  /** For logging purposes */
+  String getLabel();
+
+  boolean applies(FqlService fqlService, MigratableQueryInformation migratableQueryInformation);
```
mweaver-ebsco, line 11: This method should have some docs around it, since it's not clear what it does

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+   * The special key "*" on the field map can be used to apply a template to all fields. This value
+   * should be a string with a %s placeholder, which will be filled in with the original field name.
+   */
+  protected abstract Map<UUID, Map<String, String>> getFieldChanges();
+
+  @Override
+  public boolean applies(FqlService fqlService, MigratableQueryInformation src) {
+    String srcVersion = "0";
+    if (src.fqlQuery() != null) {
+      Fql query = fqlService.getFql(src.fqlQuery());
```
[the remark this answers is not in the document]
ncovercash: Considered this, but I didn't want to break the case of significant FQL changes (where an old query would no longer be deserializable, for example removing a $operator)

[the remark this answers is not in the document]
mweaver-ebsco: Oh yeah, that totally makes sense :)

bvsharp:
I think the V0 -> V1 migration list looks good. Regarding the ⚠️s:

1. Would it make sense to add primary_contributor and primary_address fields to the new entity types? If those were desired in the old ones, then maybe the easiest solution for migration is to just add them to the new ones too. 
2. drv_item_holdingsrecord_instance was really just item_details without a few of the extra joins, so I think it's fine if we lose that one.

ncovercash:
> Would it make sense to add primary_contributor and primary_address fields to the new entity types? If those were desired in the old ones, then maybe the easiest solution for migration is to just add them to the new ones too.

I think they were only there because we didn't have an array solution

ncovercash:
Maybe it'd be worthwhile to discuss field removal functionality and/or some special `_removed` or whatever that we could point removed fields towards?

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+      if (query._version() != null) {
+        srcVersion = query._version();
+      }
+    }
+
+    return this.getSourceVersion().equals(srcVersion);
```
mweaver-ebsco, line 42: Is this logic backward? I'd think the strategy would apply if the versions _don't_ match

ncovercash: It's correct; the simple ones have a source A and target B; we only want to run the migration to go from A -> B. If it starts at anything other than A, other migration(s) should be done to get to A before the A->B change is made

bvsharp (approved):

kjain110 (approved):

PR #341, opened 2024-07-29
MODFQMMGR-406 MODFQMMGR-408 Refactor alias handling when flattening ETs

mweaver-ebsco:
This commit changes how we approach aliases when flattening entity
types, so that instead of having to define a sourceAlias in each column
and refer to it as ":sourceAlias" in getters, we can use the source
aliases directly (e.g., if there's a "users" source, we can reference it
with ":users"


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
mweaver-ebsco, line 1: This class is where most of the changes live. I refactored a significant portion of it

src/main/resources/entity-types/inventory/composite_item_details.json5
```diff
@@ -15,7 +15,7 @@
       id: '74ddf1a6-19e0-4d63-baf0-cd2da9a46ca4', // simple_locations
       join: {
         type: 'left join',
-        joinTo: 'items',
+        joinTo: 'items.item.item',
```
bvsharp, line 18:
```suggestion
        joinTo: 'items.item',
```

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -48,54 +51,75 @@ public EntityType getFlattenedEntityType(UUID entityTypeId, boolean doFinalRenam
...
-          String oldAlias = subSource.getAlias();
-          Pair<EntityTypeSource, List<EntityTypeColumn>> updatePair = getConvertedSourceAndColumns(flattenedSourceDefinition, subSource, source, keepOriginalAlias);
-          String newAlias = updatePair.getLeft().getAlias();
-          if (!oldAlias.equals(newAlias)) {
-            updateOtherSources(oldAlias, newAlias, flattenedSourceDefinition.getSources());
-          }
-          flattenedEntityType.addSourcesItem(updatePair.getLeft());
-          finalColumns.addAll(updatePair.getRight());
-        }
+        // Add a prefix to each column's name and idColumnName, then add em to the flattened entity typeg
```
bvsharp, line 84:
```suggestion
        // Add a prefix to each column's name and idColumnName, then add em to the flattened entity type
```

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -48,54 +51,75 @@ public EntityType getFlattenedEntityType(UUID entityTypeId, boolean doFinalRenam
...
+      // Update alias
+      String oldAlias = source.getAlias();
+      String newAlias = aliasPrefix + oldAlias;
+      renamedAliases.put(oldAlias, newAlias);
+    }
 
-    List<EntityTypeColumn> finalColumns = new ArrayList<>();
     Set<String> finalPermissions = new HashSet<>(originalEntityType.getRequiredPermissions());
+    Stream.Builder<Stream<EntityTypeColumn>> columns = Stream.<Stream<EntityTypeColumn>>builder()
+      .add(copyColumns(sourceFromParent, originalEntityType, renamedAliases));
```
mweaver-ebsco, line 73: It occurred to me that this method ends up spending quite a bit of time just creating lists, only to copy the list contents into another list. Rinse and repeat. I refactored the column handling, so that it keeps them in streams and generally doesn't bother actually dumping them into the final list until the very end, to make this a little better. It'll still probably thrash the garbage collector a bit and allocate a lot more stuff than necessary, but it should be at least a tiny bit better

ncovercash (approved):

kjain110 (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -48,54 +51,75 @@ public EntityType getFlattenedEntityType(UUID entityTypeId, boolean doFinalRenam
...
-          flattenedEntityType.addSourcesItem(updatePair.getLeft());
-          finalColumns.addAll(updatePair.getRight());
-        }
+        // Add a prefix to each column's name and idColumnName, then add em to the flattened entity typeg
+        columns.add(
+          flattenedSourceDefinition.getColumns()
+            .stream()
+            .map(col -> col
+              .name(aliasPrefix + source.getAlias() + '.' + col.getName())
+              .idColumnName(col.getIdColumnName() == null ? null : aliasPrefix + source.getAlias() + '.' + col.getIdColumnName())
```
mweaver-ebsco, line 90: Here's the fix for MODFQMMGR-406 - add the prefix to `idColumnName` in the same way we do it for column names. And by doing both right next to each other, it helps ensure that we're handling both in the same way (which is important, since ID column name needs to refer to valid columns, so when column names get updated, ID column names need the same treatment)

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -171,120 +189,71 @@ private static void getSourcesRecursively(EntityTypeSource source, Map<String, E
...
+    return originalEntityType.getColumns()
+      .stream()
+      .map(column -> {
+        EntityTypeColumn newColumn = copyColumn(column, originalEntityType);
+        // Only treat newColumn as idColumn if outer source specifies to do so
+        newColumn.isIdColumn(newColumn.getIsIdColumn() == null ? null : Boolean.TRUE.equals(newColumn.getIsIdColumn()) && (sourceFromParent == null || Boolean.TRUE.equals(sourceFromParent.getUseIdColumns())));
+        injectSourceAlias(newColumn, originalEntityType, renamedAliases, newColumn.getSourceAlias());
+        newColumn.setSourceAlias(null);
+        return newColumn;
+      });
```
mweaver-ebsco, line 254: To make it a little more efficient, I opted to keep this as 1 big single `map()` step instead of breaking it into several smaller ones. In the version with smaller lambdas, each lambda ends up referring to at least 1 method parameter, which means that the JVM has to allocate an object for each lambda in order to create a closure. With just 1 big lambda, then we only get the 1 object.

Totally a micro-optimization, but it's a simple one that doesn't harm readability, so why not ¯\\\_(ツ)_/¯

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -294,21 +263,26 @@ private EntityTypeColumn copyColumn(EntityTypeColumn column, EntityType entityTy
...
+      .join(source.getJoin() == null ? null : new EntityTypeSourceJoin()
+        .type(source.getJoin().getType())
+        .condition(source.getJoin().getCondition())
+        .joinTo(renameAliases ? renamedAliases.get(source.getJoin().getJoinTo()) : source.getJoin().getJoinTo())
+      )
+      .useIdColumns(sourceFromParent == null || Boolean.TRUE.equals(source.getUseIdColumns()));
   }
 
-  private List<EntityTypeColumn> getFilteredColumns(List<EntityTypeColumn> unfilteredColumns) {
+  private Stream<EntityTypeColumn> getFilteredColumns(Stream<EntityTypeColumn> unfilteredColumns) {
```
mweaver-ebsco, line 281: Personal opinion: We should remove this method from this class, since it's not at all related to entity type flattening. But that's out of scope of this already large PR, so I didn't bother

src/main/resources/entity-types/circulation/composite_loan_details.json5
```diff
@@ -14,7 +14,7 @@
       id: '64d7b5fb-2ead-444c-a9bd-b9db831f4132',
       join: {
         type: 'left join',
-        joinTo: 'loans',
+        joinTo: 'loans.loan',
```
mweaver-ebsco, line 17: Afaik, there are only 2 breaking changes from this PR:
- the one here, where `joinTo` has to refer to a specific source in the target entity type (I could probably fix this and maintain the old behavior, but I honestly kind of prefer this approach, since it gives a bit more power when joining)
- In some cases, field names can change. I don't think this actually happened with any of our existing entity types, but it does show up in `EntityTypeFlatteningServiceTest` with the test that flattens a composite containing a composite

src/test/java/org/folio/fqm/IntegrationTestBase.java
```diff
@@ -36,7 +36,7 @@
 public class IntegrationTestBase {
 
   private static final Logger LOG = LoggerFactory.getLogger(IntegrationTestBase.class);
-  private static final boolean IS_LOG_ENABLED = false;  // for debugging set to true!
+  private static final boolean IS_LOG_ENABLED = true;  // for debugging set to true!
```
mweaver-ebsco, line 39: Revert this!

src/test/java/org/folio/fqm/IntegrationTestBase.java
```diff
@@ -172,6 +172,9 @@ public MockResponse dispatch(@NotNull RecordedRequest recordedRequest) {
           }
           """).setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE);
       }
+      if (recordedRequest.getPath().matches("/consortia-configuration.*")) {
+        return new MockResponse().setBody("").setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE);
+      }
```
mweaver-ebsco, line 177: Totally unrelated to this PR, but it reduces test noise quite a bit

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
mweaver-ebsco: Switching-windows-and-typing-before-the-switching-animation-completes strikes again!

PR #351, opened 2024-08-07
MODFQMMGR-382:Support option to limit columns inherited from an Entit…

kjain110:


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -98,6 +98,10 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
     Stream<EntityTypeColumn> childSourceColumns = columns.build().flatMap(Function.identity());
     Stream<EntityTypeColumn> allColumns = Stream.concat(copyColumns(sourceFromParent, originalEntityType, renamedAliases), childSourceColumns);
 
+    if (sourceFromParent != null && Boolean.TRUE.equals(sourceFromParent.getEssentialOnly())) {
+      allColumns = allColumns.filter(EntityTypeColumn::getEssential);
+    }
```
mweaver-ebsco, line 103: Since the `essentialOnly` only comes into play when inheriting columns, I think this should happen inside the above `for`-loop. I think it'd go right after the `stream()` on line 83.
Something like
```java
.filter(col -> !Boolean.TRUE.equals(source.getEssentialOnly()) || col.getEssential())
```

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -279,6 +283,7 @@ private static EntityTypeSource copySource(EntityTypeSource sourceFromParent, En
         .condition(source.getJoin().getCondition())
         .joinTo(renameAliases ? renamedAliases.get(source.getJoin().getJoinTo()) : source.getJoin().getJoinTo())
       )
+      .essentialOnly(sourceFromParent == null || Boolean.TRUE.equals(source.getEssentialOnly()))
```
mweaver-ebsco, line 286: I think I'd just be lazy on this one and copy the value blindly, rather than force it to be non-null
```suggestion
      .essentialOnly(source.getEssentialOnly())
```

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
mweaver-ebsco: Is the column's `essential` property nullable? If so, then you'll need to do the `Boolean.TRUE.equals()` thing for that one, too

kjain110: Yes, it's nullable. Will be adding that  Boolean.TRUE.equals() for it as well

PR #352, opened 2024-08-08
MODFQMMGR-380: Enforce permissions for cross-tenant queries

bvsharp:
## Purpose
[MODFQMMGR-380](https://folio-org.atlassian.net/browse/MODFQMMGR-380): Enforce required permissions per tenant in an ECS environment

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -39,7 +39,10 @@
           "methods": ["GET"],
           "pathPattern": "/entity-types/{entity-type-id}/columns/{column-name}/values",
           "permissionsRequired": ["fqm.entityTypes.item.columnValues.get"],
-          "modulePermissions": ["perms.users.get"]
+          "modulePermissions": [
+            "perms.users.get",
+            "consortia.user-tenants.collection.get"
+          ]
         },
         {
           "methods": ["POST"],
@@ -56,13 +59,19 @@
           "methods": ["GET"],
           "pathPattern": "/query",
           "permissionsRequired": ["fqm.query.sync.get"],
-          "modulePermissions": ["perms.users.get"]
+          "modulePermissions": [
+            "perms.users.get",
+            "consortia.user-tenants.collection.get"
+          ]
         },
         {
           "methods": ["GET"],
           "pathPattern": "/query/{query-id}",
           "permissionsRequired": ["fqm.query.async.results.get"],
-          "modulePermissions": ["perms.users.get"]
+          "modulePermissions": [
+            "perms.users.get",
+            "consortia.user-tenants.collection.get"
+          ]
         },
         {
           "methods": ["GET"],
@@ -86,7 +95,10 @@
           "methods": ["POST"],
           "pathPattern": "/query",
           "permissionsRequired": ["fqm.query.async.post"],
-          "modulePermissions": ["perms.users.get"]
+          "modulePermissions": [
+            "perms.users.get",
+            "consortia.user-tenants.collection.get"
+          ]
         },
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -39,10 +39,8 @@
           "methods": ["GET"],
           "pathPattern": "/entity-types/{entity-type-id}/columns/{column-name}/values",
           "permissionsRequired": ["fqm.entityTypes.item.columnValues.get"],
-          "modulePermissions": [
-            "perms.users.get",
-            "consortia.user-tenants.collection.get"
-          ]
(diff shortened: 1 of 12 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -56,13 +59,19 @@
...
+          ]
         },
         {
           "methods": ["GET"],
           "pathPattern": "/query/{query-id}",
           "permissionsRequired": ["fqm.query.async.results.get"],
-          "modulePermissions": ["perms.users.get"]
+          "modulePermissions": [
+            "perms.users.get",
+            "consortia.user-tenants.collection.get"
```
ncovercash, line 73: Are we going to explode anything if this is deployed in non-ECS?

ncovercash (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -22,39 +23,61 @@ public class CrossTenantQueryService {
...
     String centralTenantId;
     String consortiumId;
+    List<HashMap<String, String>> userTenantHashMaps;
     // Below if-block is necessary to limit cross-tenant querying to instance entity type until we have a way to
     // configure cross-tenant queries by entity type. This can be removed after completion of MODFQMMGR-335.
     if (!entityTypeId.equals(UUID.fromString("6b08439b-4f8e-4468-8046-ea620f5cfb74"))) {
       return List.of(executionContext.getTenantId());
     }
+
+    EntityType entityType = entityTypeFlatteningService.getFlattenedEntityType(entityTypeId, null);
```
mweaver-ebsco, line 39: Nothing actionable here yet, but one of us is going to get a merge conflict 🙂 On my branch, I changed the method signature to take an `EntityType` instead of an ID, so that'll be fun to deal with 😀

src/main/java/org/folio/fqm/service/PermissionsService.java
```diff
@@ -32,14 +32,22 @@ public class PermissionsService {
...
+    "fqm.entityTypes.item.get",
+    "fqm.query.async.results.get",
+    "fqm.query.async.post"
+  );
 
-  public Set<String> getUserPermissions() {
-    TenantUserPair key = new TenantUserPair(context.getTenantId(), context.getUserId());
-    return cache.get(key, k -> isEureka ? getUserPermissionsFromRolesKeycloak(k.userId()) : getUserPermissionsFromModPermissions(k.userId()));
+  public Set<String> getUserPermissions(String tenantId) {
+    String tenantIdForQuery = tenantId != null ? tenantId : context.getTenantId();
```
mweaver-ebsco, line 42: If we have a way to handle null tenant IDs, why not just always get it from the context instead of using a parameter? If we do need to pass it in, I'd probably overload the method so that we have a version without the tenant ID parameter instead of having to explicitly pass in `null`

kjain110 (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -22,39 +23,61 @@ public class CrossTenantQueryService {
 
   private final SimpleHttpClient ecsClient;
   private final FolioExecutionContext executionContext;
+  private final EntityTypeFlatteningService entityTypeFlatteningService;
+  private final PermissionsService permissionsService;
 
   public List<String> getTenantsToQuery(UUID entityTypeId) {
     String centralTenantId;
     String consortiumId;
+    List<HashMap<String, String>> userTenantHashMaps;
```
mweaver-ebsco, line 32: Nit: The fact that this uses `HashMap` is an implementation detail that isn't particularly important, so I wouldn't mention it in the variable name. I'd lean toward something more like `userTenantMaps`

Nit #2: Can we change the type on this to `List<Map<String, String>>`?

Nit #3 (maybe not a nit, since I care more about this one): It's not obvious what this variable is or what it contains (something about users and tenants, but I'm not sure what each map is or what the keys and values are). A short comment or something would go a long way here

descriptors/ModuleDescriptor-template.json
bvsharp: Good question, I'm honestly not sure how it would handle that ECS permission in non-ECS environments, I couldn't find much documentation on it. We do need that permission for ECS environments though. Maybe it makes more sense to put it in permissionsDesired, that way it should be ok with the permission not being present.

src/main/java/org/folio/fqm/service/PermissionsService.java
bvsharp: We have to pass the tenant ID, because it won't always come from the execution context. The execution context here will correspond to the central tenant, but the tenant ID needed here is for the other ECS tenants. I can implement it with an overloaded method though.

PR #355, opened 2024-08-08
fixing instance status for migration stuff

kjain110:


src/main/java/org/folio/fqm/migration/strategies/V0POCMigration.java
```diff
@@ -93,7 +93,6 @@ public class V0POCMigration extends AbstractSimpleMigrationStrategy {
     Map.entry("instance_statistical_code_ids", "instance.statistical_code_ids"),
     Map.entry("instance_statistical_codes", "instance.statistical_code_names"),
     Map.entry("instance_status_id", "instance.status_id"),
-    Map.entry("instance_status", "instance.status"),
```
ncovercash, line 96: This should be something like `inst_stat.name`, shouldn't it?

ncovercash: We still need to migrate the field name, just need to change it to the new one since `status` is not part of `instance` ET anymore

ncovercash (approved):

bvsharp (approved):

mweaver-ebsco (approved):

PR #359, opened 2024-08-09
MODFQMMGR-378 Add support for enabling cross-tenant queries for ETs

mweaver-ebsco:
This commit changes how we approach configuring available values for columns, so that we no longer need to depend on columns having special names. Instead, each FQM-provided set of values is configured in the column's `source` property. Additional properties within that property's model allow us to differentiate `source` configurations where the data comes from an entity type vs internal FQM logic

## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -23,14 +24,12 @@ public class CrossTenantQueryService {
   private final SimpleHttpClient ecsClient;
   private final FolioExecutionContext executionContext;
 
-  public List<String> getTenantsToQuery(UUID entityTypeId) {
```
mweaver-ebsco, line 26: If you watch the logs when doing just about anything in FQM, you'll notice that they are just FLOODED with log messages about loading the same entity types from the DB over and over again. I think a huge part of that is that we tend to pass IDs around, so I went ahead and made this change to cut down on that at least a little. Caching entity type definitions will make a much bigger difference, but this change still seemed worthwhile

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -76,6 +77,11 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
         UUID sourceEntityTypeId = UUID.fromString(source.getId());
         // Recursively flatten the source and add it to the flattened entity type
         EntityType flattenedSourceDefinition = getFlattenedEntityType(sourceEntityTypeId, source, tenantId);
+        // If the original entity type already supports cross-tenant queries, we can skip this. Otherwise, copy the nested source's setting
+        // This effectively means that if any nested source supports cross-tenant queries, the flattened entity type will too
+        if (!Boolean.TRUE.equals(flattenedEntityType.getCrossTenantQueriesEnabled())) {
+          flattenedEntityType.crossTenantQueriesEnabled(flattenedSourceDefinition.getCrossTenantQueriesEnabled());
+        }
```
mweaver-ebsco, line 84: This logic is kind of roundabout. It's not _awful_ but I do have to mentally run through every scenario for it to really make sense.

1. If `flattenedEntityType` is ECS-enabled, then it will just stay enabled.
2. If it isn't enabled and the nested source also isn't, it stays disabled.
3. If it doesn't have it enabled and the nested source _does_, then this will make `flattenedEntityType` ECS-enabled.
4. If it then runs into a nested source where it's disabled, it will automatically get skipped because `flattenedEntityType` is already ECS-enabled, so it falls into scenario 1

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -112,34 +107,24 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
...
+        switch (Objects.requireNonNull(field.getSource().getName(), "Value sources with the FQM type require the source name to be configured")) {
+          case "currency" -> {
+            return getCurrencyValues();
+          }
+          case "tenant_id" -> {
+            return getTenantIds(entityType);
+          }
+          case "source_tenant_id" -> {
+            return getSourceTenantIds(entityType);
+          }
```
mweaver-ebsco, line 123: If this list gets much larger, we'll probably want to move this `switch` block into its own method (or maybe into a `Map<String, Supplier<ColumnValues>>`

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -112,34 +107,24 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
...
-        return new ColumnValues().content(tenantValues);
-      } else {
-        // tenant is not central tenant, but we need to provide central tenant id as an available value
-        String centralTenantId = crossTenantQueryService.getCentralTenantId();
-        List<ValueWithLabel> tenantList = new ArrayList<>();
-        if (centralTenantId != null && !centralTenantId.equals(tenants.get(0))) {
-          tenantList.add(new ValueWithLabel().value(centralTenantId).label(centralTenantId));
+    if (field.getSource() != null) {
+      if (field.getSource().getType() == SourceColumn.TypeEnum.ENTITY_TYPE) {
+        return getFieldValuesFromEntityType(entityType, fieldName, searchText);
```
mweaver-ebsco, line 112: This technically introduces a regression here, but I think it's okay.

Before, this method checked the various ways to configure the available values for a column, then just fell back to `getFieldValuesFromEntityType()`. With this change, it only does `getFieldValuesFromEntityType()` if `field.getSource()` is actually set on the column, and falls back to an exception.

I think that's okay because `getFieldValuesFromEntityType()` is very far from guaranteed to work with all columns (it depends on the entity type specifically having an "id" column), and this method only ever actually gets called when the column `source` is non-null.

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -151,11 +136,36 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
       return getFieldValuesFromApi(field, searchText);
     }
 
-    if (field.getName().equals("pol_currency")) {
-      return getCurrencyValues();
+    throw new InvalidEntityTypeDefinitionException("Unable to retrieve column values for " + fieldName, entityType);
+  }
+
+  private ColumnValues getSourceTenantIds(EntityType entityType) {
```
mweaver-ebsco, line 142: This method and `getTenantIds()` are both just what we already had, but split out into new methods

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -112,34 +107,24 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
...
-        String centralTenantId = crossTenantQueryService.getCentralTenantId();
-        List<ValueWithLabel> tenantList = new ArrayList<>();
-        if (centralTenantId != null && !centralTenantId.equals(tenants.get(0))) {
-          tenantList.add(new ValueWithLabel().value(centralTenantId).label(centralTenantId));
+    if (field.getSource() != null) {
+      if (field.getSource().getType() == SourceColumn.TypeEnum.ENTITY_TYPE) {
+        return getFieldValuesFromEntityType(entityType, fieldName, searchText);
+      } else if (field.getSource().getType() == SourceColumn.TypeEnum.FQM) {
+        switch (Objects.requireNonNull(field.getSource().getName(), "Value sources with the FQM type require the source name to be configured")) {
+          case "currency" -> {
```
mweaver-ebsco, line 115: Before, currencies were handled by looking for a column named `pol_currency`, which doesn't work since we renamed that column. Now, since it uses the `name` from `source`, the column name is irrelevant, so went with `currency`

src/main/resources/entity-types/orders/simple_purchase_order_line.json5
```diff
@@ -379,8 +379,8 @@
       filterValueGetter: "lower(:sourceAlias.jsonb -> 'cost' ->> 'currency')",
       valueFunction: 'lower(:value)',
       source: {
-        entityTypeId: '58148257-bfb0-4687-8c42-d2833d772f3e',
-        columnName: 'cost_currency',
+        name: 'fqm',
+        columnName: 'currency',
```
mweaver-ebsco, line 383:
```suggestion
        entityTypeId: '58148257-bfb0-4687-8c42-d2833d772f3e',
        columnName: 'cost_currency',
        name: 'fqm',
        columnName: 'currency',
```

src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -1068,6 +1069,8 @@
       source: {
         columnName: 'source_tenant_id',
         entityTypeId: '8fc4a9d2-7ccf-4233-afb8-796911839862',
+        type: 'fqm',
```
mweaver-ebsco, line 1072: Note: the other available `type` is `entity-type`. I made that the default, though, which is why this PR doesn't modify every single `source` in the project

src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -3,6 +3,7 @@
   name: 'simple_instance',
   root: true,
   private: true,
+  crossTenantQueriesEnabled: true,
```
mweaver-ebsco, line 6: I should move this to the composite instance ET. Otherwise, it will get inherited by composite_items and anything else that uses it, which could be confusing

ncovercash (approved):

PR #366, opened 2024-08-13
MODFQMMGR-424: Remove shadow instances from results

bvsharp:
## Purpose
[MODFQMMGR-424](https://folio-org.atlassian.net/browse/MODFQMMGR-424): Remove shadow instances from results

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -24,6 +24,10 @@ public class CrossTenantQueryService {
   private final FolioExecutionContext executionContext;
   private final PermissionsService permissionsService;
 
+  private static final String CROSS_TENANT_QUERY_ERROR = "Error retrieving tenants for cross-tenant query. Tenant may not be in an ECS environment.";
+  private static final String CONSORTIA_CONFIGURATION_PATH = "consortia-configuration";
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -79,12 +83,24 @@ public List<String> getTenantsToQuery(EntityType entityType) {
...
+    try {
+      String rawJson = ecsClient.get(CONSORTIA_CONFIGURATION_PATH, Map.of());
+      DocumentContext parsedJson = JsonPath.parse(rawJson);
+      parsedJson.read(CENTRAL_TENANT_ID);
+      return true;
+    } catch (Exception e) {
+      log.debug(CROSS_TENANT_QUERY_ERROR);
+      return false;
+    }
+  }
```
bvsharp, line 105: This essentially duplicates a method we have in the EntityTypeFlatteningService, but it makes more sense to have here. I was going to replace the method in EntityTypeFlatteningService with a call to this one, but that leads to a circular dependency that would be more of a headache than it's worth dealing with. So I think for now it's probably ok to have it in both spots, though this is something I'd want to clean up in the future.

ncovercash (approved):

kjain110 (approved):

PR #370, opened 2024-08-14
Adding an end point /fqm/migrate

kjain110:
Screen Shot 2024-09-09 at 4 21 13 PM

### DIFF
--- a/src/main/java/org/folio/fqm/resource/MigrationController.java
+++ b/src/main/java/org/folio/fqm/resource/MigrationController.java
@@ -1,18 +1,53 @@
 package org.folio.fqm.resource;
 
 import lombok.RequiredArgsConstructor;
+import org.folio.fqm.domain.dto.FqmMigrateRequest;
+import org.folio.fqm.domain.dto.FqmMigrateResponse;
+import org.folio.fqm.domain.dto.FqmMigrateResponseWarningsInner;
+import org.folio.fqm.migration.MigratableQueryInformation;
 import org.folio.fqm.service.MigrationService;
+import org.folio.spring.i18n.service.TranslationService;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
+
 
 @RestController
 @RequiredArgsConstructor
 public class MigrationController implements FqmVersionApi {
   private final MigrationService migrationService;
 
+  private final TranslationService translationService;
+
   @Override
   public ResponseEntity<String> getFqmVersion() {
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/fqm/resource/MigrationController.java
```diff
@@ -15,4 +19,24 @@ public class MigrationController implements FqmVersionApi {
...
+      fqmMigrateRequest.getFields()
+    );
+
+    migrationService.migrate(migratableQueryInformation);
+
+    FqmMigrateResponse fqmMigrateResponse = new FqmMigrateResponse()
+      .entityTypeId(migratableQueryInformation.entityTypeId())
+      .fqlQuery(migratableQueryInformation.fqlQuery())
+      .fields(migratableQueryInformation.fields())
+      .warnings(migratableQueryInformation.warnings().stream()
```
ncovercash, line 37: I'd like to return the type and description in the API response, so the description can be returned to the user and the type can be used by the receiving module/etc to handle the case appropriately

src/main/resources/swagger.api/schemas/FqmMigrateResponse.json
```diff
@@ -0,0 +1,34 @@
...
+        "type": "string"
+      }
+    },
+  "warnings": {
+    "description": "List of warnings issued during the upgrade",
+    "type": "array",
+    "items": {
+      "type": "string"
+    }
+  }
```
ncovercash, line 29: see above comment
```suggestion
  "warnings": {
    "description": "List of warnings issued during the upgrade",
    "type": "array",
    "items": {
      "type": "object",
      "properties": {
        "description": { "type": "string" },
        "type": { "type": "string" }
      }
    }
  }
```

ncovercash (approved):

src/main/resources/swagger.api/mod-fqm-manager.yaml
```diff
@@ -79,6 +79,30 @@ paths:
...
+        - fqmVersion
+      requestBody:
+        description: 'Request for FQM version submitted successfully'
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/fqmMigrateRequest'
+      responses:
+        '201':
```
ncovercash, line 96: Probably 200

ncovercash (approved):

src/main/resources/swagger.api/mod-fqm-manager.yaml
kjain110: Nice catch! Done

PR #377, opened 2024-08-20
MODFQMMGR-415 Add an entity type defintion cache

mweaver-ebsco:
This commit adds a new cache for entity types to EntityTypeRepository, to cut down on repeated loads of entity types. The default cache duration is 1 hour, but this can be overridden via system property. Additionally, the value in the module descriptor is set to 5 minutes, in order to strike more of a balance between performance and practicality in dev environments (although devs should consider reducing this further when actively creating/updating entity type definitions)

## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -405,6 +405,9 @@
       },
       { "name": "MAX_QUERY_SIZE",
         "value": "1250000"
+      },
+      { "name": "mod.fqm-manager.entity-type-cache-timeout-seconds",
+        "value": "300"
       }
     ]
   }
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -1,7 +1,8 @@
 package org.folio.fqm.repository;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.RequiredArgsConstructor;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import lombok.SneakyThrows;
 import lombok.extern.log4j.Log4j2;
 import org.folio.fqm.exception.EntityTypeNotFoundException;
@@ -9,26 +10,24 @@ import org.folio.querytool.domain.dto.BooleanType;
 import org.folio.querytool.domain.dto.EntityType;
 import org.folio.querytool.domain.dto.EntityTypeColumn;
 import org.folio.querytool.domain.dto.ValueWithLabel;
-import org.jooq.Condition;
-import org.jooq.DSLContext;
-import org.jooq.Field;
-import org.jooq.InsertValuesStep2;
-import org.jooq.JSONB;
 import org.jooq.Record;
+import org.jooq.*;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Repository;
 
 import java.util.*;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.apache.commons.collections4.CollectionUtils.isEmpty;
 import static org.jooq.impl.DSL.field;
-import static org.jooq.impl.DSL.or;
 import static org.jooq.impl.DSL.table;
-import static org.jooq.impl.DSL.trueCondition;
 
 @Repository
-@RequiredArgsConstructor
 @Log4j2
 public class EntityTypeRepository {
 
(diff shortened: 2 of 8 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -405,6 +405,9 @@
       },
       { "name": "MAX_QUERY_SIZE",
         "value": "1250000"
+      },
+      { "name": "mod.fqm-manager.entity-type-cache-timeout-seconds",
+        "value": "300"
```
mweaver-ebsco, line 410: This seems high enough to make an impact, but still low enough to not make updating entity types frustrating.

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -48,30 +53,45 @@ public class EntityTypeRepository {
...
 
+  @Value("${mod-fqm-manager.entity-type-cache-timeout-seconds:3600}")
+  private long cacheDurationSeconds;
+  private final Cache<String, Map<UUID, EntityType>> entityTypeCache = Caffeine.newBuilder().expireAfterWrite(cacheDurationSeconds, TimeUnit.SECONDS).build();
+
   public Optional<EntityType> getEntityTypeDefinition(UUID entityTypeId, String tenantId) {
     return getEntityTypeDefinitions(Collections.singleton(entityTypeId), tenantId).findFirst();
   }
 
   public Stream<EntityType> getEntityTypeDefinitions(Collection<UUID> entityTypeIds, String tenantId) {
```
mweaver-ebsco, line 64: The major change in this method is that instead of only loading some ETs in some cases and all of them in others, it just always loads them all. Then, all entity type defs are just served up from the cache

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -98,10 +118,10 @@ public void replaceEntityTypeDefinitions(List<EntityType> entityTypes) {
     });
   }
 
-  private List<EntityTypeColumn> fetchColumnNamesForCustomFields(UUID entityTypeId, EntityType entityType) {
+  private List<EntityTypeColumn> fetchColumnNamesForCustomFields(String entityTypeId, EntityType entityType, Map<String, EntityType> rawEntityTypes) {
     log.info("Getting columns for entity type ID: {}", entityTypeId);
-    EntityType entityTypeDefinition = entityTypeId.toString().equals(entityType.getId()) ? entityType :
-      getEntityTypeDefinition(entityTypeId, null).orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+    EntityType entityTypeDefinition = entityTypeId.equals(entityType.getId()) ? entityType :
+      Optional.ofNullable(rawEntityTypes.get(entityTypeId)).orElseThrow(() -> new EntityTypeNotFoundException(UUID.fromString(entityTypeId)));
```
mweaver-ebsco, line 124: I had to change this, since the previous approach would make it try to load from the cache while it's still trying to populate the cache, which causes an exception

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -48,30 +53,45 @@ public class EntityTypeRepository {
...
-      .map(this::unmarshallEntityType)
-      .map(entityType -> {
-        String customFieldsEntityTypeId = entityType.getCustomFieldEntityTypeId();
-        if (customFieldsEntityTypeId != null) {
-          entityType.getColumns().addAll(fetchColumnNamesForCustomFields(UUID.fromString(customFieldsEntityTypeId), entityType));
-        }
-        return entityType;
-      });
+    Map<UUID, EntityType> entityTypes = entityTypeCache.get(tenantId == null ? "" : tenantId, tenantIdKey -> {
+        String tableName = "".equals(tenantIdKey) ? TABLE_NAME : tenantIdKey + "_mod_fqm_manager." + TABLE_NAME;
```
mweaver-ebsco, line 68: This `""` change is kind of hacky, but you can't have a `null` cache key, and turning `null` into a special value is super easy and doesn't feel quite as gross when the special value replacement is done on the line above where it's handled

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -48,30 +53,45 @@ public class EntityTypeRepository {
...
+        Field<String> definitionField = field(DEFINITION_FIELD_NAME, String.class);
+      Map<String, EntityType> rawEntityTypes = readerJooqContext
+        .select(definitionField)
+        .from(table(tableName))
+        .where(trueCondition())
```
mweaver-ebsco, line 73: @mweaver-ebsco, we don't need a `WHERE` clause anymore, dummy!

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -143,5 +163,7 @@ private EntityType unmarshallEntityType(String str) {
     return objectMapper.readValue(str, EntityType.class);
   }
 
-  public record RawEntityTypeSummary(UUID id, String name, List<String> requiredPermissions) {}
+  public void clearCache() {
+    entityTypeCache.invalidateAll();
+  }
```
mweaver-ebsco, line 168: Dead code

src/test/java/org/folio/fqm/utils/IdStreamerTestDataProvider.java
```diff
@@ -66,7 +66,7 @@ public class IdStreamerTestDataProvider implements MockDataProvider {
         .target("target1"))
     );
 
-  private static final String ENTITY_TYPE_DEFINITION_REGEX = "SELECT DEFINITION FROM .*ENTITY_TYPE_DEFINITION WHERE ID IN .*";
+  private static final String ENTITY_TYPE_DEFINITION_REGEX = "SELECT DEFINITION FROM .*ENTITY_TYPE_DEFINITION WHERE .*";
```
mweaver-ebsco, line 69: When the `WHERE` clause goes away, this will need to be updated

descriptors/ModuleDescriptor-template.json
ncovercash: The cache still gets nuked on update, yeah? If we want 300 to be the default I'd sync the readme/application.properties, too

ncovercash (approved):

bvsharp (approved):

descriptors/ModuleDescriptor-template.json
mweaver-ebsco: Yep, anything that restarts the JVM (or even that nukes the `EntityTypeRepository` instance) would clear the cache. I think 3600 (1 hour) feels like a reasonable default for production. That's a terrible default for dev envs, though. I was thinking 300 strikes a nice balance where it's okay for dev without being awful for production (still not great for prod, but at least it's not awful)

mweaver-ebsco: I went ahead and added a paragraph to the readme to make it clear that the value in the module descriptor is different from the default, along with my reasoning. I also added some simple recommendations for production (basically: higher timeouts = good. 1 day or 1 week would be solid choices)

PR #378, opened 2024-08-20
MODFQMMGR-439 Mark all simples in composite_instances as non-essentia…

mweaver-ebsco:
…l-only

## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist


ncovercash (approved):

mweaver-ebsco:
For context: removing the non-essential columns is making life difficult for our PO, so we're just doing this temporarily. These will almost certainly need to be switched back to `essentialOnly` soon, but that's easy enough

PR #380, opened 2024-08-21
MODFQMMGR-425: Retrieve shared instances when in a member tenant

bvsharp:
## Purpose
[MODFQMMGR-425](https://folio-org.atlassian.net/browse/MODFQMMGR-425): Retrieve shared instances when in a member tenant

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -27,6 +27,7 @@ public class CrossTenantQueryService {
   private static final String CROSS_TENANT_QUERY_ERROR = "Error retrieving tenants for cross-tenant query. Tenant may not be in an ECS environment.";
   private static final String CONSORTIA_CONFIGURATION_PATH = "consortia-configuration";
   private static final String CENTRAL_TENANT_ID = "centralTenantId";
+  private static final String COMPOSITE_INSTANCES_ID = "6b08439b-4f8e-4468-8046-ea620f5cfb74";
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -48,6 +49,9 @@ public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossT
           .read("consortia[0].id");
       } else {
         log.debug("Tenant {} is not central tenant. Running intra-tenant query.", executionContext.getTenantId());
+        if (COMPOSITE_INSTANCES_ID.equals(entityType.getId())) {
```
mweaver-ebsco, line 52: We should probably create a ticket to replace this with an entity type property of some kind, since it feels kind of gross to hard-code the ID

mweaver-ebsco: Also, a comment here to explain the weirdness would be super useful later on when reading this code

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
bvsharp: We could, but it's such a specific thing that seems like it will only apply to composite instances. I'm not aware of anything else that needs to retrieve shared records from the central tenant, so to me it feels like the property would clutter things more (especially since we already have a bunch that rarely get used).

kjain110 (approved):

PR #382, opened 2024-08-21
MODFQMMGR-421 Add tenant_id to items and holdings

mweaver-ebsco:


src/main/resources/entity-types/inventory/simple_holdings_records.json5
```diff
@@ -836,6 +836,25 @@
...
+      visibleByDefault: true,
+      valueGetter: "'${tenant_id}'",
+      source: {
+        columnName: 'tenant_id',
+        entityTypeId: '30a5cfad-1868-4f46-86b9-a6ef67e2d9bf',
+        type: 'fqm',
+        name: 'tenant_id'
+      },
+      ecsOnly: false,
+    },
```
mweaver-ebsco, line 857: Literally copy/paste from instances 🙂  All I changed is the `source.entityTypeId`

ncovercash (approved):

kjain110 (approved):

src/main/resources/entity-types/inventory/simple_holdings_records.json5
```diff
@@ -836,6 +836,25 @@
       visibleByDefault: false,
       valueGetter: ":sourceAlias.jsonb->>'notes'",
     },
+    {
+      name: 'tenant_id',
+      sourceAlias: 'inst',
```
bvsharp, line 841: It does feel a little weird to have 
sourceAlias: 'inst' 
here, but since there isn't really a source alias for this field, and since I think things break if we don't at least have something defined for the field, it's probably ok.

bvsharp (approved):

src/main/resources/entity-types/inventory/simple_holdings_records.json5
mweaver-ebsco: Oh yeah :D Good catch!

PR #383, opened 2024-08-22
MODFQMMGR-444:Holdings - permanent location returns no results

kjain110:


ncovercash (approved):

src/main/resources/entity-types/inventory/simple_locations.json5
```diff
@@ -63,8 +63,8 @@
       queryable: true,
       visibleByDefault: false,
       valueGetter: ":sourceAlias.jsonb ->> 'name'",
-      filterValueGetter: "lower(${tenant_id}_mod_inventory_storage.f_unaccent(:sourceAlias.jsonb ->> 'name'::text))",
-      valueFunction: 'lower(${tenant_id}_mod_inventory_storage.f_unaccent(:value))',
+      filterValueGetter: ":sourceAlias.id::text",
```
mweaver-ebsco, line 66: Um, are you sure this is right for the name?

kjain110: Yes, otherwise we are comparing name to id. In order to make things work we need :(

mweaver-ebsco: I'm confused... Why would we be comparing an ID to this name column when the name column doesn't have an `idColumnName`?

PR #385, opened 2024-08-22
MODFQMMGR-448 Fix bug with valueSourceApi

mweaver-ebsco:
With a recent change, the source property on columns started taking precedence over valueSourceApi, but source is required on columns that use valueSourceApi, making it effectively break all valueSourceApi columns. This commit rearranges taht code, so that valueSourceApi (and values, which is more of an edge case, but has the same bug) has a higher precedence

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -115,6 +115,14 @@ public class EntityTypeService {
       .findFieldDefinition(new FqlField(fieldName), entityType)
       .orElseThrow(() -> new FieldNotFoundException(entityType.getName(), fieldName));
 
+    if (field.getValues() != null) {
+      return getFieldValuesFromEntityTypeDefinition(field, searchText);
+    }
+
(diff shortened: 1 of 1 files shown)


ncovercash (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -115,6 +115,14 @@ public ColumnValues getFieldValues(UUID entityTypeId, String fieldName, @Nullabl
...
+    if (field.getValues() != null) {
+      return getFieldValuesFromEntityTypeDefinition(field, searchText);
+    }
+
+    if (field.getValueSourceApi() != null) {
+      return getFieldValuesFromApi(field, searchText);
+    }
+
     if (field.getSource() != null) {
       if (field.getSource().getType() == SourceColumn.TypeEnum.ENTITY_TYPE) {
```
mweaver-ebsco, line 127: If the source type isn't defined, it defaults to entity-type, so it was always ending up trying to look up values using the DB instead of the API. Whoops!

PR #393, opened 2024-08-26
MODFQMMGR-420 Fix bug where column names sometimes had repeated source aliases

mweaver-ebsco:
This commit fixes a bug where composite entity types containing other composite entity types ended up containing columns with names that were confusingly named and longer than necessary, since some of the source alias from the ET's hierarchy were repeated in the column name.


ncovercash (approved):
pls update translations as applicable

bvsharp (approved):

PR #396, opened 2024-08-27
[MODFQMMGR-453] Add local development permissions bypass

ncovercash:
Looks worse than it is; moved `PermissionsService` into `PermissionsRegularService` with virtually zero difference


src/main/java/org/folio/fqm/config/PermissionsConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Log4j2
+@Configuration
+public class PermissionsConfiguration {
```
ncovercash, line 17: this is where the real spring ✨magic✨ happens

bvsharp (approved):

mweaver-ebsco (approved):

PR #401, opened 2024-08-29
[MODFQMMGR-427] Fix mode of issuance column

ncovercash:
Also:
- added a tool to the entity type creator to make querying values easier to test
- sorted the translations file (sorry for the diff!)


src/main/resources/entity-types/inventory/simple_mode_of_issuance.json5
```diff
@@ -0,0 +1,47 @@
...
+      isIdColumn: true,
+      queryable: true,
+      visibleByDefault: false,
+      valueGetter: ':sourceAlias.id',
+    },
+    {
+      name: 'name',
+      sourceAlias: 'mode_of_issuance',
+      dataType: {
+        dataType: 'stringType',
```
kjain110, line 30: we don't require valueSourceApi thing for the drop down ?

ncovercash: that's only if it's from an API, this one comes from DB

kjain110 (approved):

PR #405, opened 2024-08-30
MODFQMMGR-409 Prepend the ET name to column names

mweaver-ebsco:
This commit makes FQM prepend the root entity type's (possibly shortened) name to the name of every column in that entity type


src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -65,6 +67,14 @@ public void localizeEntityTypeColumn(EntityType entityType, EntityTypeColumn col
...
       column.setLabelAlias(sourceTranslation + column.getLabelAlias());
     }
+    if (prependShortenedName) {
+      String untranslated = ENTITY_TYPE_SHORTENED_LABEL_TRANSLATION_TEMPLATE.formatted(entityType.getName());
+      String shortenedName = translationService.format(ENTITY_TYPE_SHORTENED_LABEL_TRANSLATION_TEMPLATE.formatted(entityType.getName()));
+      if (shortenedName.equals(untranslated)) {
+        shortenedName = getEntityTypeLabel(entityType.getName());
+      }
+      column.setLabelAlias(shortenedName + " — " + column.getLabelAlias());
+    }
```
mweaver-ebsco, line 77: Need at least 1 unit test for this (2 or 3 would be better, so that we get full branch coverage)

src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -78,7 +88,7 @@ private String getSourceTranslationPrefix(EntityType entityType, String columnNa
     }
   }
 
-  protected void localizeObjectColumn(EntityType entityType, EntityTypeColumn column, ObjectType objectColumn) {
+  private void localizeObjectColumn(EntityType entityType, EntityTypeColumn column, ObjectType objectColumn) {
```
mweaver-ebsco, line 91: The visibility on these methods was bothering me, so I went ahead and locked them down as much as I could

ncovercash (approved):

PR #406, opened 2024-08-30
[MODFQMMGR-429] Expose resource type JSONB

ncovercash:


src/main/resources/entity-types/inventory/simple_instance_type.json5
```diff
@@ -136,6 +136,7 @@
       isIdColumn: false,
       queryable: false,
       hidden: true,
+      essential: true,
```
kjain110, line 139: why do we need this ?

ncovercash: ticket said we wanted _only_ jsonb

kjain110 (approved):

PR #408, opened 2024-09-03
[MODFQMMGR-249] Hide instance_type_id field

ncovercash:
Take 3 🙃

### DIFF
--- a/src/main/resources/entity-types/inventory/simple_instance.json5
+++ b/src/main/resources/entity-types/inventory/simple_instance.json5
@@ -118,6 +118,7 @@
       dataType: {
         dataType: 'rangedUUIDType',
       },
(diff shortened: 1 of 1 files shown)


mweaver-ebsco (approved):

src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -118,6 +118,7 @@
       dataType: {
         dataType: 'rangedUUIDType',
       },
+      hidden: true,
       isIdColumn: false,
       queryable: true,
       essential: false,
```
mweaver-ebsco, line 124: We might need to make this essential, but that can/should probably happen in a different ticket

PR #412, opened 2024-09-04
Quick fix to ensure custom fields work with user and POL functionality

kjain110:


ncovercash (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -90,17 +90,23 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
             .stream()
             .filter(col -> !Boolean.TRUE.equals(source.getEssentialOnly()) || Boolean.TRUE.equals(col.getEssential()))
             .map(col -> col
-              // Don't use aliasPrefix here, since the prefix is already appropriately baked into the source aliass in flattenedSourceDefinition
```
mweaver-ebsco, line 93: This comment seems useful enough to keep around, since it explains something that may not be obvious

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -109,6 +115,20 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
     return localizationService.localizeEntityType(flattenedEntityType, sourceFromParent == null);
   }
 
+    private String injectSourceAliasIntoViewExtractor(String sourceViewExtractor, Map<String, String> renamedAliases) {
+    // Convert the keySet to a list and sort it in natural order, then reverse it
+    List<String> aliases = new ArrayList<>(renamedAliases.keySet());
+    aliases.sort(Comparator.naturalOrder());  // Ensure natural order first
```
mweaver-ebsco, line 121: I don't think we should sort the list here. `renamedAliases` is a `LinkedHashMap` so it should already be sorted as a pre-order traversal of the entity type hierarchy, so I wouldn't touch the ordering at all

PR #413, opened 2024-09-05
MODFQMMGR-462 Get tenant data from user-tenants

mweaver-ebsco:


ncovercash (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -1,6 +1,5 @@
 package org.folio.fqm.service;
```
mweaver-ebsco, line 1: The major changes are in this class

bvsharp: This is a lot nicer!

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -88,26 +61,31 @@ public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossT
...
-      DocumentContext parsedJson = JsonPath.parse(rawJson);
-      return parsedJson.read(CENTRAL_TENANT_ID);
+      String userTenantsResponse = ecsClient.get("user-tenants", Map.of());
+      return JsonPath
+          .parse(userTenantsResponse)
+          .read("$.userTenants", List.class);
     } catch (Exception e) {
-      log.debug(CROSS_TENANT_QUERY_ERROR);
-      return null;
+      log.error(CROSS_TENANT_QUERY_ERROR, e);
```
bvsharp, line 72: nit (feel free to ignore): I'd keep this as debug to avoid cluttering the logs with lots of ERRORs when running queries in non-ECS environments. Since for any non-ECS environment, you'd be seeing this message every time even though there's not really an error running the query.

bvsharp (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -88,26 +61,31 @@ public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossT
...
   }
 
-  public String getCentralTenantId() {
+  @SuppressWarnings("unchecked") // JsonPath.parse is returning a plain List without a type parameter, and the TypeRef (vs Class) parameter to JsonPath.read is not supported by the JSON parser
+  private List<Map<String, String>> getUserTenants() {
     try {
-      String rawJson = ecsClient.get(CONSORTIA_CONFIGURATION_PATH, Map.of());
-      DocumentContext parsedJson = JsonPath.parse(rawJson);
-      return parsedJson.read(CENTRAL_TENANT_ID);
+      String userTenantsResponse = ecsClient.get("user-tenants", Map.of());
```
bvsharp, line 67:
```suggestion
      String userTenantsResponse = ecsClient.get("user-tenants", Map.of("limit", String.valueOf(1000)));
```

I think this should resolve the problem pointed out in https://folio-org.atlassian.net/browse/MODFQMMGR-465 (at least up to 1000 tenants)

ncovercash (approved):

kjain110 (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
mweaver-ebsco: Honestly, I should probably remove the `try` block here, since exceptions would be truly exceptional. We should still get a response in non-ECS envs

bvsharp: Oh yeah good point, I was thinking of the old logic

bvsharp (approved):

PR #421, opened 2024-09-13
MODFQMMGR-462 Add a missing module permission + logging

mweaver-ebsco:
## Purpose

## Approach

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -1,6 +1,7 @@
 package org.folio.fqm.service;
 
 import com.jayway.jsonpath.JsonPath;
+import feign.FeignException;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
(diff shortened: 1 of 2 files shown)


kjain110 (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -54,6 +55,9 @@ public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossT
           tenantsToQuery.add(tenantId);
         } catch (MissingPermissionsException e) {
           log.info("User with id {} does not have permissions to query tenant {}. Skipping.", userId, tenantId);
+        } catch (FeignException e) {
+          log.error("Error retrieving permissions for user ID %s in tenant %s".formatted(userId, tenantId), e);
```
mweaver-ebsco, line 59: Log this, to provide a little more useful info that isn't in the exception that shows up right now (then re-throw it, since we don't want to fail silently here)

PR #426, opened 2024-09-17
MODFQMMGR-477 Fix bug with user-tenants API call

mweaver-ebsco:
This commit fixes a bug where FQM was retrieving user tenants incorrectly, resulting in it attempting to retrieve user permissions from tenants where the user may not exist

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -13,7 +13,6 @@ import org.springframework.stereotype.Service;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 @Service
 @RequiredArgsConstructor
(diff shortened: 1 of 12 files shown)


bvsharp (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -68,7 +68,7 @@ public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossT
   @SuppressWarnings("unchecked")
   // JsonPath.parse is returning a plain List without a type parameter, and the TypeRef (vs Class) parameter to JsonPath.read is not supported by the JSON parser
   private List<Map<String, String>> getUserTenants() {
-    String userTenantsResponse = ecsClient.get("user-tenants", Map.of("limit", "1000"));
+    String userTenantsResponse = ecsClient.get("user-tenants", Map.of("userId", executionContext.getUserId().toString(), "limit", "1000"));
```
bvsharp, line 71: I probably should have noticed this was missing when I reviewed the previous one... 😬

kjain110 (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
mweaver-ebsco: Turns out it shouldn't be there anyway 😛 In member tenants, there's only a dummy user with a random userId, so adding a `userId` query param makes it filter the dummy user out. However, we need that dummy user to get the central tenant ID and consortium ID

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -25,27 +24,34 @@ public class CrossTenantQueryService {
...
 
   public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossTenantQuery) {
     if (!forceCrossTenantQuery && !Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())) {
       return List.of(executionContext.getTenantId());
     }
-    List<Map<String, String>> userTenantMaps = getUserTenants();
-    String centralTenantId = getCentralTenantId(userTenantMaps); // null if non-ECS environment; non-null otherwise
+    // Get the user's primary affiliation first, since this comes from mod-users and should work in non-ECS environments
+    // We can use this for determining if it's an ECS environment, and if so, retrieving the consortium ID and central tenant ID
+    Map<String, String> primaryAffiliation = getPrimaryAffiliation();
```
mweaver-ebsco, line 35: This should probably all get reworded/renamed, since it's not actually getting the user's primary affiliation anymore

ncovercash (approved):

bvsharp (approved):

PR #429, opened 2024-09-19
MODFQMMGR-484: Use index for simple_instance.source field

bvsharp:
## Purpose
[MODFQMMGR-484](https://folio-org.atlassian.net/browse/MODFQMMGR-484): Use index for simple_instance.source field

### DIFF
--- a/src/main/resources/entity-types/inventory/simple_instance.json5
+++ b/src/main/resources/entity-types/inventory/simple_instance.json5
@@ -206,7 +206,6 @@
     },
     {
       name: 'source',
-      sourceAlias: 'inst',
       dataType: {
         dataType: 'stringType',
       },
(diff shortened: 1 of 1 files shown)


ncovercash (approved):

src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -214,7 +214,8 @@
       queryable: true,
       visibleByDefault: true,
       essential: true,
-      valueGetter: ":sourceAlias.jsonb->>'source'",
```
ncovercash, line 217: Do we still need the sourceAlias prop, then?

ncovercash (approved):

src/main/resources/entity-types/inventory/simple_instance.json5
mweaver-ebsco: Nope. Only needed if it's used

mweaver-ebsco (approved):

PR #431, opened 2024-09-23
MODFQMMGR-273:Restrict query execution based on entity-type subpermis…

kjain110:


src/main/resources/application.yml
```diff
@@ -3,7 +3,7 @@ mod-fqm-manager:
   max-query-size: ${MAX_QUERY_SIZE:1250000}
   bypass-permissions: false # local development ONLY
   permissions-cache-timeout-seconds: 60
-  entity-type-cache-timeout-seconds: 3600
+  entity-type-cache-timeout-seconds: 3
```
mweaver-ebsco, line 6: This change probably shouldn't get committed :)

src/main/resources/application.yml
```diff
@@ -76,7 +76,7 @@ coffee-boots:
     spec:
       queryCache: maximumSize=500,expireAfterWrite=1m
 folio:
-  is-eureka: false
+  is-eureka: true
```
mweaver-ebsco, line 79: This one probably shouldn't get committed either

src/test/java/org/folio/fqm/service/PermissionsRegularServiceTest.java
```diff
@@ -7,40 +7,43 @@
...
-  }
-
   void setUpMocks(String... permissions) {
     var userId = UUID.randomUUID();
     when(context.getUserId()).thenReturn(userId);
     when(context.getTenantId()).thenReturn(TENANT_ID);
-    when(modPermissionsClient.getPermissionsForUser(TENANT_ID, userId.toString()))
-      .thenReturn(new ModPermissionsClient.UserPermissions(List.of(permissions), permissions.length));
+
+    if (permissionsService.isEureka) {
```
ncovercash, line 37: it feels like we should have tests for both cases...

kjain110: then it would go in the else block right or you mean something else

ncovercash (changes requested):
see comments

src/test/java/org/folio/fqm/service/PermissionsRegularServiceTest.java
ncovercash: i mean we'd need one test with `isEureka = true` and another `isEureka = false`

src/main/java/org/folio/fqm/service/PermissionsRegularService.java
```diff
@@ -27,36 +27,43 @@
...
 
+  @Override
   public Set<String> getUserPermissions() {
     return getUserPermissions(context.getTenantId());
   }
 
   public Set<String> getUserPermissions(String tenantId) {
     TenantUserPair key = new TenantUserPair(tenantId, context.getUserId());
-    return cache.get(key, k -> isEureka ? getUserPermissionsFromRolesKeycloak(k.userId()) : getUserPermissionsFromModPermissions(tenantId, k.userId()));
+    return cache.get(key, k -> isEureka ? getUserPermissionsFromRolesKeycloak(tenantId, k.userId()) : getUserPermissionsFromModPermissions(tenantId, k.userId()));
```
mweaver-ebsco, line 58: Nit: Get the tenant ID from `k` instead of using `tenantId`. That way, the lambda function doesn't have to be instantiated every time the method is called (if you reference something outside the lambda, then the lambda's has to be a closure that covers whatever scope the thing lives in. So by referencing `isEureka`, the lambda is effectively an instance variable that and only needs to be instantiated once per instance. By referencing a method parameter, the closure has a method scope, so it needs instantiated for every method call)

src/main/java/org/folio/fqm/client/ModRolesKeyClockClient.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+@Service
+@FeignClient(name = "permissions")
+
+public interface ModRolesKeyClockClient {
```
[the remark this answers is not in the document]
kjain110: I hope this makes you happy @mweaver-ebsco  :)

src/test/java/org/folio/fqm/service/PermissionsRegularServiceTest.java
kjain110: I hope the update makes you happy @ncovercash :)

src/main/java/org/folio/fqm/service/PermissionsRegularService.java
kjain110: Done

src/test/java/org/folio/fqm/service/PermissionsRegularServiceTest.java
```diff
@@ -134,9 +137,31 @@ void shouldCheckFqmPermissionsIfRequested() {
...
+  void WhenEurekaIsFalse() {
+    permissionsService.isEureka = false;
+    setUpMocks("permission1", "permission2");
+    Set<String> userPermissions = permissionsService.getUserPermissions();
+
+    ArgumentCaptor<String> userIdCaptor = ArgumentCaptor.forClass(String.class);
+    verify(modPermissionsClient).getPermissionsForUser(eq(TENANT_ID), userIdCaptor.capture());
+    assertEquals(2, userPermissions.size());
+
+    // Ensure no interactions with modRolesKeyclockClient;
```
ncovercash, line 163: stupid nit: please remove the `;` to make sonar happy (it thinks this is a line of code)
```suggestion
    // Ensure no interactions with modRolesKeyclockClient
```

bvsharp (approved):

PR #434, opened 2024-09-24
MODFQMMGR-456: Check whether entity type is cross-tenant when retrieving definition

bvsharp:
## Purpose
[MODFQMMGR-456](https://folio-org.atlassian.net/browse/MODFQMMGR-456): Check whether entity type is cross-tenant when retrieving definition

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -22,12 +22,15 @@ public class CrossTenantQueryService {
   private final SimpleHttpClient ecsClient;
   private final FolioExecutionContext executionContext;
   private final PermissionsService permissionsService;
+  private final UserTenantService userTenantService;
 
   private static final String COMPOSITE_INSTANCES_ID = "6b08439b-4f8e-4468-8046-ea620f5cfb74";
   private static final String SIMPLE_INSTANCES_ID = "8fc4a9d2-7ccf-4233-afb8-796911839862";
 
   public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossTenantQuery) {
-    if (!forceCrossTenantQuery && !Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())) {
+    if (!forceCrossTenantQuery
+      && !Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())
+      && !COMPOSITE_INSTANCES_ID.equals(entityType.getId())) {
       return List.of(executionContext.getTenantId());
     }
     // Get the ECS tenant info first, since this comes from mod-users and should work in non-ECS environments
@@ -82,6 +85,23 @@ public class CrossTenantQueryService {
       .parse(userTenantResponse)
       .read("$.userTenants", List.class);
   }
+
+  public String getCentralTenantId() {
+    return getCentralTenantId(getEcsTenantInfo());
+  }
+
+  public boolean ecsEnabled() {
+    return ecsEnabled(getEcsTenantInfo());
+  }
+
+  public boolean isCentralTenant() {
+    return isCentralTenant(getEcsTenantInfo());
+  }
+
+  private boolean ecsEnabled(Map<String, String> ecsTenantInfo) {
+    return !(ecsTenantInfo == null || ecsTenantInfo.isEmpty());
+  }
+
   /**
    * Retrieve the primary affiliation for a user.
    * This retrieves the primary affiliation for an arbitrary user in the tenant.
@@ -90,7 +110,7 @@ public class CrossTenantQueryService {
    */
   @SuppressWarnings("unchecked") // JsonPath.parse is returning a plain List without a type parameter, and the TypeRef (vs Class) parameter to JsonPath.read is not supported by the JSON parser
   private Map<String, String> getEcsTenantInfo() {
-    String userTenantsResponse = ecsClient.get("user-tenants", Map.of("limit", "1"));
+    String userTenantsResponse = userTenantService.getUserTenantsResponse(executionContext.getTenantId());
     List<Map<String, String>> userTenants = JsonPath
       .parse(userTenantsResponse)
       .read("$.userTenants", List.class);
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -27,7 +27,9 @@ public class CrossTenantQueryService {
   private static final String SIMPLE_INSTANCES_ID = "8fc4a9d2-7ccf-4233-afb8-796911839862";
 
   public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossTenantQuery) {
-    if (!forceCrossTenantQuery && !Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())) {
+    if (!forceCrossTenantQuery
+      && !Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())
+      && !COMPOSITE_INSTANCES_ID.equals(entityType.getId())) {
```
bvsharp, line 32: I hate how complicated this condition is getting, but we do need all these checks. The new check for the composite instances ID is due to the fact that cross-tenant queries will be disabled for non-central tenants, but we DO still need to run a (pseudo) cross tenant query for composite instances in member tenants in order to get the shared instances from the central tenant.

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -86,6 +87,8 @@ public List<EntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds, boo
    */
   public EntityType getEntityTypeDefinition(UUID entityTypeId, boolean includeHidden, boolean sortColumns) {
     EntityType entityType = entityTypeFlatteningService.getFlattenedEntityType(entityTypeId, null);
+    boolean crossTenantEnabled = Boolean.TRUE.equals(entityType.getCrossTenantQueriesEnabled())
+      && crossTenantQueryService.isCentralTenant();
```
bvsharp, line 91: Kind of a weird way to do it, but this essentially overwrites the `crossTenantQueriesEnabled` property, so that non-central tenants have it set to false, even if it's true in the original definition. This will allow us to avoid forcing lists for cross-tenant entity types to be private in member tenants, and instead only apply that effect to lists in the central tenant. Still not crazy about this approach but it seems to be the simplest option for now.

ncovercash (approved):

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
mweaver-ebsco: This is going to run into issues with the tenant ID dropdown menu for instances, since the entity type that gets passed in is simple_instances, which doesn't have cross-tenant queries enabled :(

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -115,4 +115,12 @@ private boolean ecsEnabled(Map<String, String> ecsTenantInfo) {
   public boolean ecsEnabled() {
     return ecsEnabled(getEcsTenantInfo());
   }
+
+  public boolean isCentralTenant() {
+    return executionContext.getTenantId().equals(getCentralTenantId());
+  }
```
mweaver-ebsco, line 121: I'd follow the pattern that the other method here use, where there's overloaded with a `private` version that takes the `Map<String, String> ecsTenantInfo` parameter. Then, I'd update `getTenantsToQuery()` to use the `private` version, so that all of the "is this the central tenant?" logic runs through that one method

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
mweaver-ebsco: If we end up needing to check both composite and simple instances in 2 places in this method, I think I'd make a `private static final Set<String>` to hold the IDs, then do a `contains(entityType.getId())` instead of checking both IDs separately

bvsharp: This situation actually gets handled by the `forceCrossTenantQuery` parameter! When we get the tenant ID values, we pass `forceCrossTenantQuery=true` to this method. Since it's true, we skip this block. From there:
  1. If ECS is not enabled, we return only the current tenant.
  2. If ECS is enabled but we're not in the central tenant AND the entity type is composite/simple instances, we return the current tenant plus the central tenant. 
  3. If ECS is enabled and we're in the central tenant, then we query everything we have permission for. 

So I think everything related to the tenant ID dropdown is handled. The only thing that the current code doesn't handle is running a (non-tenant ID related) query on the simple_instances entity from a member tenant. In that situation, we'd only be querying the member tenant, so we wouldn't see the shared records from the central tenant (although this is counter-balanced by the fact that simple_instances doesn't have an `additionalEcsConditions` defined either, meaning that we wouldn't be filtering out shadow instances. So in this situation we'd see ALMOST the same thing as for composite_instances, only we'd be seeing the shadow instances instead of the shared ones).

bvsharp: And I don't think we'd want to include simple_instances in the first check (at least the way we have the definitions set up right now), due to the fact the simple_instances doesn't filter out the shadow instances. So if we added the simple_instance check to the first block, we'd end up with duplicate instance records in the query results (one shadow version and one shared version for each shared instance).

src/main/java/org/folio/fqm/service/UserTenantService.java
```diff
@@ -0,0 +1,26 @@
...
+public class UserTenantService {
+
+  private final SimpleHttpClient userTenantsClient;
+
+  @Cacheable(value="userTenantCache", key="#tenantId")
+  public String getUserTenantsResponse(String tenantId) {
+    log.info("Retrieving user-tenants information for tenant {}", tenantId);
+    return userTenantsClient.get("user-tenants", Map.of("limit", String.valueOf(1)));
+  }
+}
```
bvsharp, line 26: Added this class to allow for convenient caching for the user-tenants responses. We use the /user-tenants endpoint in both the EntityTypeFlatteningService and the CrossTenantQueryService, so I figured adding a service wrapper made the most sense to allow us to:
  1. Cache the user-tenants responses in a convenient way regardless of what class the request is made from.
  2. Avoid caching responses on any of the other SimpleHttpClient requests.

src/main/resources/application.yml
```diff
@@ -75,6 +75,7 @@ coffee-boots:
   cache:
     spec:
       queryCache: maximumSize=500,expireAfterWrite=1m
+      userTenantCache: maximumSize=100,expireAfterWrite=5h
```
bvsharp, line 78: Picked a fairly arbitrary expire time, and am open to any other values for this cache. Also open to making it a configurable value, though I didn't do make it configurable for the time being.

src/main/java/org/folio/fqm/service/UserTenantService.java
mweaver-ebsco, line 1: Good call on breaking this out into a separate service!

mweaver-ebsco (approved):

kjain110 (approved):

PR #436, opened 2024-09-26
MODFQMMGR-471: Instance-contributors

kjain110:


ncovercash (commented):
Please add/change migration scripts as needed!

src/main/java/org/folio/fqm/migration/strategies/V0POCMigration.java
```diff
@@ -442,14 +442,14 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
           "instance_primary_contributor",
-          (String field, String fql) -> new QueryBreakingWarning(field, "instance.contributors", fql)
+          (String field, String fql) -> new QueryBreakingWarning(field, "instance.instance_primary_contributor", fql)
         )
       ),
       Map.entry(
         OLD_DRV_ITEM_DETAILS,
         Map.of(
           "instance_primary_contributor",
-          (String field, String fql) -> new QueryBreakingWarning(field, "instance.contributors", fql)
```
ncovercash, line 452: Since the primary field has returned, you can entirely remove this Warning, and update the map higher up in this file

ncovercash (changes requested):

ncovercash (approved):

PR #450, opened 2024-10-04
[MODFQMMGR-468] Aggregate Tenant locations across all tenants

bvsharp:
## Purpose
[MODFQMMGR-468](https://folio-org.atlassian.net/browse/MODFQMMGR-468): Aggregate Tenant locations across all tenants

### DIFF
--- a/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
+++ b/src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
@@ -26,13 +26,35 @@ public class CrossTenantQueryService {
 
   private static final String COMPOSITE_INSTANCES_ID = "6b08439b-4f8e-4468-8046-ea620f5cfb74";
   private static final String SIMPLE_INSTANCES_ID = "8fc4a9d2-7ccf-4233-afb8-796911839862";
+  private static final String SIMPLE_INSTANCE_STATUS_ID = "9c239bfd-198f-4013-bbc4-4551c0cbdeaa";
+  private static final String SIMPLE_INSTANCE_TYPE_ID = "af44e2e0-12e0-4eec-b80d-49feb33a866c";
+  private static final List<String> INSTANCE_RELATED_ENTITIES = List.of(SIMPLE_INSTANCES_ID, COMPOSITE_INSTANCES_ID, SIMPLE_INSTANCE_STATUS_ID, SIMPLE_INSTANCE_TYPE_ID);
 
-  public List<String> getTenantsToQuery(EntityType entityType, boolean forceCrossTenantQuery) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -26,13 +26,35 @@ public class CrossTenantQueryService {
...
+   * column values API uses simple entity type definitions, which don't have cross-tenant queries enabled.
+   * method skips the cross-tenant query check
+   * @param entityType Entity type definition
+   * @return           List of tenants to query
+   */
+  public List<String> getTenantsToQueryForColumnValues(EntityType entityType) {
+    return getTenants(entityType);
+  }
+
+  private List<String> getTenants(EntityType entityType) {
```
bvsharp, line 57: I split the old method into 2 just to try to make the conditional logic easier to follow. `getTenantsToQuery` is used to get tenants when actually running a query, while `getTenantsToQueryForColumnValues` is used just for the column values API. This way made the most sense to me because the query builder uses simple entity type definitions when making the requests for column values. These simple definitions don't have cross-tenant queries enabled, so we need to bypass the `crossTenantQueriesEnabled` check when getting the tenants to query for column values.

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -26,13 +26,35 @@ public class CrossTenantQueryService {
 
   private static final String COMPOSITE_INSTANCES_ID = "6b08439b-4f8e-4468-8046-ea620f5cfb74";
   private static final String SIMPLE_INSTANCES_ID = "8fc4a9d2-7ccf-4233-afb8-796911839862";
+  private static final String SIMPLE_INSTANCE_STATUS_ID = "9c239bfd-198f-4013-bbc4-4551c0cbdeaa";
+  private static final String SIMPLE_INSTANCE_TYPE_ID = "af44e2e0-12e0-4eec-b80d-49feb33a866c";
+  private static final List<String> INSTANCE_RELATED_ENTITIES = List.of(SIMPLE_INSTANCES_ID, COMPOSITE_INSTANCES_ID, SIMPLE_INSTANCE_STATUS_ID, SIMPLE_INSTANCE_TYPE_ID);
```
bvsharp, line 31: Contains all entity types related to composite_instances. We use this list when retrieving column values, so that we are able to provide cross-tenant values for all composite instance fields that use a dropdown (even though the simple entity type itself is not cross-tenant).

mweaver-ebsco (approved):

kjain110 (approved):

PR #453, opened 2024-10-07
MODFQMMGR-330: Add Languages field to Instance entity

bvsharp:
## Purpose
[MODFQMMGR-330](https://folio-org.atlassian.net/browse/MODFQMMGR-330): Add Languages field to Instance entity. Also removes all materialized views other than statistical codes


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -222,6 +246,70 @@ private static ColumnValues getCurrencyValues() {
...
+    try(InputStream input = getClass().getClassLoader().getResourceAsStream(LANGUAGES_FILEPATH)) {
+      languages = mapper.readValue(input, new TypeReference<>() {
+      });
+    } catch (IOException e) {
+      log.error("Failed to read language file. Language display names may not be properly translated.");
+    }
+
+    Locale folioLocale;
+    try {
+      String localeSettingsResponse = simpleHttpClient.get(GET_LOCALE_SETTINGS_PATH, GET_LOCALE_SETTINGS_PARAMS);
```
ncovercash, line 270: are we unable to make this a feign client? :(

PR #478, opened 2024-10-16
MODFQMMGR-497 Change data type for instance dates

mweaver-ebsco:


src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -691,7 +691,7 @@
       sourceAlias: 'inst',
       essential: true,
       dataType: {
-        dataType: 'dateType',
+        dataType: 'stringType',
```
mweaver-ebsco, line 694: It turns out, these values are really just unstructured strings ¯\\\_(ツ)_/¯

kjain110 (approved):

ncovercash (approved):

PR #495, opened 2024-10-21
Simple users entity for users of FOLIO

kjain110:


bvsharp (approved):

src/main/resources/entity-types/orders/composite_purchase_order_lines.json5
```diff
@@ -24,7 +24,7 @@
     {
       alias: 'po_created_by_user',
       type: 'entity-type',
-      id: 'bb058933-cd06-4539-bd3a-6f248ff98ee2', // simple_user_details,
+      id: 'f2615ea6-450b-425d-804d-6a495afd9308', // simple_FOIO_user_details,
```
mweaver-ebsco, line 27: Typo (5 times 😛)
```suggestion
      id: 'f2615ea6-450b-425d-804d-6a495afd9308', // simple_FOLIO_user_details,
```

src/main/resources/entity-types/users/simple_FOLIO_user_details.json5
mweaver-ebsco, line 1: Any chance we could make `simple_user_details` inherit this ET so that we don't duplicate fields?

src/main/resources/entity-types/orders/composite_purchase_order_lines.json5
kjain110: ahh, copy paste cons :)

src/main/resources/entity-types/users/simple_FOLIO_user_details.json5
```diff
@@ -0,0 +1,154 @@
+{
+  id: 'f2615ea6-450b-425d-804d-6a495afd9308',
+  name: 'simple_FOLIO_user_details',
```
mweaver-ebsco, line 3: Can we give this a more descriptive name? At a glance, it's not at all obvious what the difference between this and `simple_user_details` based on the names

kjain110: @mweaver-ebsco , or this PR can we have as it is:
https://github.com/folio-org/mod-fqm-manager/pull/495. Since if we are removing duplicate column, then all the flags needs to be reconsidered for other ET's as well. And kathleen wants it before freeze. I can create another for addressing your thoughts.

PR #508, opened 2024-10-25
[MODFQMMGR-409] Remove composite entity names from field labels, for real this time

ncovercash:


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -39,7 +39,7 @@ public EntityType getFlattenedEntityType(UUID entityTypeId, String tenantId) {
   }
 
   private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource sourceFromParent, String tenantId) {
-  // TODO: clean up
+    // TODO: clean up
```
ncovercash, line 42: I cleaned it up :)

bvsharp: Appreciate you taking the time to handle this massive piece of tech debt.

bvsharp (approved):

mweaver-ebsco (approved):

PR #515, opened 2024-10-29
MODFQMMGR-417:Mark the /query/{id}/sortedIds API endpoint as deprecated

kjain110:


ncovercash (approved):

ncovercash (approved):

ncovercash (changes requested):
~~Please deprecate via module descriptor~~

src/main/java/org/folio/fqm/resource/FqlQueryController.java
```diff
@@ -44,6 +45,13 @@ public ResponseEntity<List<Map<String, Object>>> getContents(ContentsRequest con
       contentsRequest.getFields(), contentsRequest.getIds()));
   }
 
+  /**
+   * This endpoint is deprecated and will be removed in a future version.
+   * @deprecated Since version 3.0.0.
+   * This endpoint is planned for removal.
+   */
+  @Deprecated(since = "3.0.0", forRemoval = true)
```
mweaver-ebsco, line 53: Can we mark this as deprecated in the OpenAPI spec, too (in folio-query-tool-metadata)?

ncovercash (approved):

src/main/java/org/folio/fqm/resource/FqlQueryController.java
kjain110: @mweaver-ebsco  https://github.com/folio-org/folio-query-tool-metadata/pull/71

bvsharp (approved):

PR #519, opened 2024-10-30
Modfqmmgr 558

kjain110:


src/main/resources/entity-types/organizations/simple_organizations.json5
```diff
@@ -1306,27 +1306,50 @@
...
       visibleByDefault: false,
-      valueGetter: ":sourceAlias.jsonb->>'accounts'",
+      valueGetter: "(SELECT jsonb_agg(record.value || \
+                                    jsonb_build_object('acqUnit', \
+                                    (SELECT jsonb_agg(acq_unit.jsonb ->> 'name') \
+                                     FROM jsonb_array_elements_text(record.value -> 'acqUnitIds') id \
+                                     JOIN src_acquisitions_unit acq_unit \
+                                       ON id.value::text = acq_unit.id::text)) \
+                                   ORDER BY record.value ->> 'acqUnitIds') \
+                 FROM jsonb_array_elements((:sourceAlias.jsonb -> 'accounts')) record(value))::text"
```
bvsharp, line 1352: I'm kind of confused by this, shouldn't this be the valueGetter for the entire accounts object?

kjain110: this is for the entire accounts objects. It's at the column level not at the nested column level. @bvsharp

mweaver-ebsco: I think this is doing the fun shenanigans where it adds an extra property to the object (which explains why the `property` would be "acqUnit")

kjain110: This is how we have for contributors and classification as well

mweaver-ebsco (approved):

bvsharp (approved):

PR #524, opened 2024-11-04
MODFQMMGR-544: Remove final materialized view

bvsharp:
## Purpose
[MODFQMMGR-544](https://folio-org.atlassian.net/browse/MODFQMMGR-544): Remove final materialized view


src/main/resources/db/changelog/changes/v1.1.0/sql/create-view-drv-inventory-statistical-code-full.sql
```diff
@@ -1,12 +1,11 @@
 DROP MATERIALIZED VIEW IF EXISTS drv_inventory_statistical_code_full;
+DROP MATERIALIZED VIEW IF EXISTS drv_circulation_loan_status;
+DROP MATERIALIZED VIEW IF EXISTS drv_inventory_item_status;
```
bvsharp, line 3: I noticed that these 2 were still hanging around even though we don't use them. Could've put it in its own changeset but figured it'd be fine to just throw it in here.

kjain110 (approved):
LGTM :)

PR #529, opened 2024-11-08
MODFQMMGR-532: Aggregate languages from all tenants in ECS environments

bvsharp:
## Purpose
[MODFQMMGR-532](https://folio-org.atlassian.net/browse/MODFQMMGR-532): Aggregate languages from all tenants in ECS environments

### DIFF
--- a/src/main/java/org/folio/fqm/client/CrossTenantClient.java
+++ b/src/main/java/org/folio/fqm/client/CrossTenantClient.java
@@ -1,0 +1,66 @@
+package org.folio.fqm.client;
+
+import feign.Client;
+import feign.Request;
+import feign.Response;
+import feign.okhttp.OkHttpClient;
+import lombok.extern.log4j.Log4j2;
+import org.apache.commons.lang3.StringUtils;
+import org.folio.spring.FolioExecutionContext;
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/fqm/client/CrossTenantClient.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.folio.spring.FolioExecutionContext;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+@Log4j2
+public class CrossTenantClient implements Client {
```
bvsharp, line 17: New Client implementation that's used by our LanguageClient. This is necessary in order to avoid using the default Client implementation from folio-spring-base. By using this implementation instead, the LanguageClient is able to manually set the x-okapi-tenant header, rather than using the one from the execution context (which allows us to send cross-tenant requests).

mweaver-ebsco (approved):

ncovercash (approved):

kjain110 (approved):

PR #533, opened 2024-11-12
[MODFQMMGR-573] Consider converted date for timezone offsets, rather than system time

ncovercash:
## Why did this happen?

The backend was using the current date to determine the UTC offset, which resulted in either four or five hours (in America/New_York), depending on the date that the server handled the request. This could cause some discrepancies and for queries to return different values when entering/leaving daylight savings/summer time.

### DIFF
--- a/src/main/java/org/folio/fqm/service/ResultSetService.java
+++ b/src/main/java/org/folio/fqm/service/ResultSetService.java
@@ -89,8 +89,7 @@ public class ResultSetService {
     for (String fieldName : dateFields) {
       contents.computeIfPresent(fieldName, (key, value) -> {
         if (value instanceof Timestamp ts) {
-        log.info("{} {} {}", ts, ts.toInstant(), ts.toLocalDateTime());
-        return adjustDate(ts.toInstant(), tenantTimezone);
+          return adjustDate(ts.toInstant(), tenantTimezone);
         } else if (value instanceof String s) {
           return parseAndAdjustDate(s, tenantTimezone);
         }
--- a/src/main/java/org/folio/fqm/service/ResultSetService.java
+++ b/src/main/java/org/folio/fqm/service/ResultSetService.java
@@ -1,9 +1,5 @@
 package org.folio.fqm.service;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import feign.FeignException;
 import lombok.extern.log4j.Log4j2;
 import org.folio.fqm.client.ConfigurationClient;
 import org.folio.fqm.repository.ResultSetRepository;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/fqm/service/ResultSetService.java
```diff
@@ -34,10 +30,11 @@
 @Log4j2
 public class ResultSetService {
 
-  private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-  private static final DateTimeFormatter DATE_TIME_OFFSET_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'+00:00'");
-  private static final String DATE_TIME_REGEX = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d{3}Z$";
-  private static final String DATE_TIME_OFFSET_REGEX = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d{3}[+-]\\d{2}:\\d{2}$";
+  private static final DateTimeFormatter DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
+    .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+    .optionalStart().appendOffsetId() // optional Z/timezone at end
```
ncovercash, line 35: This allows an offset at the end, or not, in both `Z` or `+HH:mm` (or any of these):
image

src/main/java/org/folio/fqm/service/ResultSetService.java
```diff
@@ -34,10 +30,11 @@
...
+    .toFormatter().withZone(ZoneOffset.UTC); // force interpretation as UTC
+  private static final String DATE_TIME_REGEX = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d{3}([+-]\\d{2}:\\d{2}(:\\d{2})?|Z|)$";
```
ncovercash, line 37: bad news: technically, timezone offsets can have a seconds place.

anyways, this combines the previous regex into one, since both can get parsed by the same formatter. 
For convenience:
image

src/test/java/org/folio/fqm/service/ResultSetServiceTest.java
```diff
@@ -43,19 +47,7 @@ class ResultSetServiceTest {
...
     );
-  private static final String LOCALE_SETTINGS = """
-    {
-      "configs": [
-        {
-           "id":"2a132a01-623b-4d3a-9d9a-2feb777665c2",
-           "module":"ORG",
-           "configName":"localeSettings",
-           "enabled":true,
-           "value":"{\\"locale\\":\\"en-US\\",\\"timezone\\":\\"Africa/Lagos\\",\\"currency\\":\\"USD\\"}"
```
ncovercash, line 54: Lagos is fortunate enough to not use summer time, so we use Ceuta instead

kjain110 (approved):

PR #543, opened 2024-11-29
MODFQMMGR-583:Use available index for defined fields

kjain110:


ncovercash (approved):

src/main/resources/entity-types/inventory/simple_holdings_records.json5
```diff
@@ -214,7 +214,9 @@
       queryable: true,
       visibleByDefault: false,
       valueGetter: ":sourceAlias.jsonb->>'callNumber'",
-      essential: true
+      essential: true,
+      filterValueGetter: "\"left\"(lower(:sourceAlias.jsonb ->> 'call_number'::text), 600)",
```
bvsharp, line 218: The valueGetter and filterValueGetter are using different jsonb fields here, which seems incorrect

src/main/resources/entity-types/inventory/simple_service_point_details.json5
```diff
@@ -140,6 +140,8 @@
           label: 'False',
         },
       ],
+      filterValueGetter: "\"left\"(lower(:sourceAlias.jsonb ->> 'pickupLocation'::text), 600)",
+      valueFunction: "\"left\"(lower(:value), 600)"
```
mweaver-ebsco, line 144: Nit: Can you put these next to the `valueGetter`? Having `valueGetter`, `filterValueGetter`, and `valueFunction` grouped together helps with readability, since the 3 are closely related

PR #546, opened 2024-12-04
Release v3.0.4

mweaver-ebsco:


bvsharp (approved):

ncovercash (approved):

NEWS.md
```diff
@@ -1,5 +1,12 @@
 # 3.0.x - Ramsons
 
+## 3.0.4
+- [MODFQMMGR-577]: Reduce the default cache expiration time for the entity type cache
```
ncovercash, line 4: Nit: * not -

ncovercash (changes requested):

NEWS.md
mweaver-ebsco: (summary of side conversation in chat)

All of the other lists in the file use `-`. We should probably change the `*`s to `-`s to make things consistent. But that's a pain, since the v3.0.4 tag is already created, so let's just do it later ¯\\\_(ツ)\_/¯

PR #548, opened 2024-12-05
[MODFQMMGR-510] Cancel queries in db

bvsharp:
## Purpose
[MODFQMMGR-510](https://folio-org.atlassian.net/browse/MODFQMMGR-510): Test query could not be Canceled from UI


src/main/java/org/folio/fqm/repository/IdStreamer.java
```diff
@@ -44,23 +46,26 @@
 @Log4j2
 public class IdStreamer {
 
+  private static final long QUERY_CANCELLATION_CHECK_SECONDS = 30;
```
bvsharp, line 49: Open to other values here. Maybe 60s?

ncovercash (approved):

kjain110 (approved):

mweaver-ebsco (approved):

PR #549, opened 2024-12-06
[MODFQMMGR-588] Migration scripts update for extra ramsons cleanup

ncovercash:


src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -36,20 +36,26 @@ public abstract class AbstractSimpleMigrationStrategy implements MigrationStrate
   public abstract String getTargetVersion();
 
   /** The entity types that got a new UUID */
-  public abstract Map<UUID, UUID> getEntityTypeChanges();
+  public Map<UUID, UUID> getEntityTypeChanges() {
+    return Map.of();
+  }
```
ncovercash, line 41: all these have defaults now

src/main/java/org/folio/fqm/migration/strategies/V0POCMigration.java
```diff
@@ -411,27 +410,27 @@ public Map<UUID, Function<String, EntityTypeWarning>> getEntityTypeWarnings() {
     return Map.ofEntries(
       Map.entry(
         UUID.fromString("146dfba5-cdc9-45f5-a8a1-3fdc454c9ae2"),
-        fql -> new RemovedEntityWarning("drv_loan_status", "simple_loans", fql)
+        RemovedEntityWarning.withAlternative("drv_loan_status", "simple_loans")
```
ncovercash, line 413: I refactored all the warnings to use factory methods instead, as having to redeclare a function to pass things around each time was kinda icky

docs/Migration.md
```diff
@@ -0,0 +1,185 @@
+# Query Migration
```
ncovercash, line 1: no excuse not to add migration scripts now 👀

bvsharp (approved):

PR #550, opened 2024-12-10
MODFQMMGR-591 Make API calls for user permissions use CrossTenantFeig…

mweaver-ebsco:
…nConfig


ncovercash (approved):

dmtkachenko:
There are three more places that use strings instead of constants, could you please also make replacements there

image

PR #552, opened 2024-12-12
[MODFQMMGR-599] Remove $ne operators from UUID field queries

ncovercash:
Updated the [wiki page](https://teams.microsoft.com/l/message/19:342d0b2d8f5d4500bf8f09499da53925@thread.v2/1734040723793?context=%7B%22contextType%22%3A%22chat%22%7D) for the poppy migration to note that those changes are being accounted for, too.


src/main/java/org/folio/fqm/migration/strategies/V5UUIDNotEqualOperatorRemoval.java
```diff
@@ -0,0 +1,238 @@
...
+                  // everything else is left behind
+                  .fql(new ObjectMapper().valueToTree(Map.of($NE, conditions.get($NE))).toPrettyString())
+                  .build()
+              );
+              conditions.remove($NE);
+            }
+
+            if (conditions.size() != 0) {
+              result.set(key, conditions);
+            }
```
ncovercash, line 232: actual migration logic

bvsharp (approved):

PR #555, opened 2024-12-16
Modfqmmgr 548:Refactor array handling to treat arrays as JSONB

kjain110:


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -323,16 +323,56 @@ private static Condition handleNotContainsAll(NotContainsAllCondition notContain
...
+    boolean isJsonbArray = isJsonbArray(entityType, containsAnyCondition);
+
+    if (isJsonbArray) {
+      // Handle as JSONB array
+      var jsonbConditions = containsAnyCondition
+        .value()
+        .stream()
+        .map(val -> {
+          var jsonbValue = valueField(val, containsAnyCondition, entityType);
+          return String.format("(%s @> '[\"%s\"]'::jsonb)", field, jsonbValue);
```
mweaver-ebsco, line 345: Rather than do this with `String.format()`, is there a way to use SQL variable binding here to make it safer (I'm not sure what the type of `jsonbValue` is, so I have no idea what will happen when you drop it into a string like this)

ncovercash (approved):

mweaver-ebsco (approved):

bvsharp (approved):

PR #556, opened 2024-12-16
[MODFQMMGR-602] Migration infrastructure improvements

ncovercash:


descriptors/ModuleDescriptor-template.json
```diff
@@ -82,7 +82,7 @@
           "methods": ["POST"],
           "pathPattern": "/fqm/migrate",
           "permissionsRequired": ["fqm.migrate.post"],
-          "modulePermissions": ["perms.users.get", "permissions.users.item.get"]
+          "modulePermissions": ["perms.users.get", "permissions.users.item.get", "configuration.entries.collection.get", "inventory-storage.modes-of-issuance.collection.get"]
```
ncovercash, line 85: mod-configuration is needed for migration V4 (adding times to dates); MOIs will be for this one

src/main/java/org/folio/fqm/migration/strategies/V5UUIDNotEqualOperatorRemoval.java
```diff
@@ -198,7 +198,7 @@ public boolean applies(String version) {
 
   @Override
   public MigratableQueryInformation apply(FqlService fqlService, MigratableQueryInformation query) {
-    Set<String> fieldsToTarget = UUID_FIELDS.get(query.entityTypeId());
+    Set<String> fieldsToTarget = UUID_FIELDS.getOrDefault(query.entityTypeId(), Set.of());
```
ncovercash, line 201: oops

src/test/java/org/folio/fqm/migration/strategies/TestTemplate.java
```diff
@@ -25,14 +27,31 @@ public abstract class TestTemplate {
...
+      .map(args -> {
+        if (args.get().length == 3) {
+          return Arguments.of(args.get()[0], args.get()[1], args.get()[2], null);
+        } else {
+          return args;
+        }
+      })
+      .toList();
+  }
+
```
ncovercash, line 42: added an optional parameter to call extra assertions after migrating (mostly for mockito)

src/test/java/org/folio/fqm/migration/strategies/V4DateFieldTimezoneAdditionTest.java
```diff
@@ -33,6 +35,22 @@ public MigrationStrategy getStrategy() {
...
+          .fqlQuery("{\"name\": {\"$eq\":\"foo\"}}")
+          .fields(List.of())
+          .build(),
+        MigratableQueryInformation
+          .builder()
+          .entityTypeId(UUID.fromString("3ad4b672-a69c-5a80-b483-bbc77c29cbfd"))
+          .fqlQuery("{\"name\": {\"$eq\":\"foo\"}, \"_version\": \"5\"}")
+          .fields(List.of())
+          .build(),
+        (Consumer<MigratableQueryInformation>) transformed -> verifyNoInteractions(configurationClient)
```
ncovercash, line 52: aforementioned extra mockito assertions, in action

src/main/java/org/folio/fqm/migration/MigrationUtils.java
```diff
@@ -101,4 +103,113 @@ private static ObjectNode migrateFqlTree(ObjectNode fql, TriConsumer<ObjectNode,
...
+                    textValue,
+                    () -> objectMapper.createObjectNode().set(entry.getKey(), entry.getValue()).toPrettyString()
+                  )
+              );
+
+              if (transformedValue != null) {
+                newValues.set(entry.getKey(), transformedValue);
+              }
+            }
+          });
```
ncovercash, line 183: this is a hot mess, primarily because we need to consider arrays used for inputs (`$in` and friends), so there's a fair bit that looks the same (but with one changed parameter).

The idea is this utility function can handle iteration through all the json nodes, pulling values out of conditions/arrays/etc, and all migration scripts need to do is tell it how to transform (add timezone, use new value, etc etc)

mweaver-ebsco (approved):

kjain110 (approved):

bvsharp (approved):

PR #562, opened 2024-12-19
Set Hikari max pool size to 12

bvsharp:
If we have 12 application threads, we need a hikari pool size of 12 too.

Also fix module descriptor

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -465,8 +465,8 @@
       { "name": "QUERY_RETENTION_DURATION",
         "value": "3h"
       },
-      { "name": "task.execution.pool.max-size",
-        "value": "10"
+      { "name": "spring.task.execution.pool.max-size",
+        "value": "12"
       },
       { "name": "MAX_QUERY_SIZE",
         "value": "1250000"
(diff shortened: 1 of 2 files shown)


kjain110 (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -465,8 +465,8 @@
       { "name": "QUERY_RETENTION_DURATION",
         "value": "3h"
       },
-      { "name": "task.execution.pool.max-size",
-        "value": "10"
+      { "name": "spring.task.execution.pool.max-size",
```
mweaver-ebsco, line 468: Should this match the change in `application.yml`? If so, then this should probably be `spring.datasource.hikari.maximum-pool-size`?

ncovercash (approved):

mweaver-ebsco (approved):

descriptors/ModuleDescriptor-template.json
bvsharp: This part is unrelated to the hikari change, and was actually just correcting a problem we've had for a while where the variable for application threads in the module descriptor didn't match what we had in application.yml. I just happened to notice that the name was wrong when working on the hikari issue. We don't allow specification of any of the hikari values in the module descriptor.

(Although your comment made me realize that we do have DB_MAXPOOLSIZE in the module descriptor. I don't think we actually use this for anything at all, so maybe it should be removed)

PR #575, opened 2025-01-07
[MODFQMMGR-625] Aggregate dropdown values from all tenants in ECS environments

bvsharp:
[[MODFQMMGR-625] Aggregate dropdown values from all tenants in ECS environments](https://github.com/folio-org/mod-fqm-manager/pull/575/commits/9317836b39947cb014c319cb34a0cd6212f84e22)

### DIFF
--- a/src/main/java/org/folio/fqm/client/CrossTenantHttpClient.java
+++ b/src/main/java/org/folio/fqm/client/CrossTenantHttpClient.java
@@ -1,24 +1,28 @@
 package org.folio.fqm.client;
 
+import org.folio.fqm.config.CrossTenantFeignConfig;
+import org.folio.spring.integration.XOkapiHeaders;
 import org.springframework.cloud.openfeign.FeignClient;
 import org.springframework.cloud.openfeign.SpringQueryMap;
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/fqm/client/SimpleHttpClient.java
```diff
@@ -12,21 +13,11 @@
...
 public interface SimpleHttpClient {
-  /**
-   * Retrieve arbitrary data from a FOLIO API endpoint.
-   *
-   * @param path        - the path of the API endpoint
-   * @param queryParams - a map of query parameters to pass to the API endpoint
-   * @return the body of the response (JSON)
-   */
-  @GetMapping(value = "/{path}", produces = MediaType.APPLICATION_JSON_VALUE)
-  String get(@PathVariable String path, @SpringQueryMap Map<String, String> queryParams);
```
ncovercash, line 25: I'm not in love with removing this (dirties up use where we don't need a custom tenant) — could we maybe re-add as `getInCurrentTenant` or something?

bvsharp: I had a feeling someone might have a problem with that 😄. Yeah I can do that

mweaver-ebsco (approved):

PR #576, opened 2025-01-10
MODFQMMGR-617:Convert specific arrays as JSONB to improve performance

kjain110:
This is a missing piece from the [MODFQMMGR-617](https://folio-org.atlassian.net/browse/MODFQMMGR-617). These two flags got missed in between refactoring of the code

### DIFF
--- a/src/main/resources/entity-types/inventory/simple_instance.json5
+++ b/src/main/resources/entity-types/inventory/simple_instance.json5
@@ -794,6 +794,8 @@
         valueJsonPath: '$.instanceFormats.*.id',
         labelJsonPath: '$.instanceFormats.*.name',
       },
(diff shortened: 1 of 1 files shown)


ncovercash (approved):

bvsharp (approved):

src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -794,6 +794,8 @@
         valueJsonPath: '$.instanceFormats.*.id',
         labelJsonPath: '$.instanceFormats.*.name',
       },
+      sourceAlias: 'inst',
```
mweaver-ebsco, line 797: This doesn't look like it's used, so you can remove it

PR #588, opened 2025-01-17
MODFQMMGR-376: Add textbox custom fields to the Users entity type

bvsharp:
## Purpose
[MODFQMMGR-376](https://folio-org.atlassian.net/browse/MODFQMMGR-376): Add textbox custom fields to the Users entity type

### DIFF
--- a/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
+++ b/src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
@@ -44,6 +44,7 @@ public class EntityTypeRepository {
 
   public static final String ID_FIELD_NAME = "id";
   public static final String DEFINITION_FIELD_NAME = "definition";
+  public static final String STRING_EXTRACTOR = "%s ->> '%s'";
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -225,7 +232,7 @@ private EntityTypeColumn handleSingleSelectCustomField(String name,
                                                          String sourceViewName,
                                                          String sourceViewExtractor,
                                                          List<ValueWithLabel> columnValues) {
-    String filterValueGetter = sourceViewExtractor + " ->> '" + refId + "'";
+    String filterValueGetter = sourceViewExtractor + STRING_EXTRACTOR + refId + "'";
```
ncovercash, line 235: nit, since you're here:

could we make STRING_EXTRACTOR something like `%s ->> '%s'` and we do `STRING_EXTRACTOR.formatted(sourceViewExtractor, refId)`?

ncovercash (approved):

mweaver-ebsco (approved):

kjain110 (approved):

PR #592, opened 2025-01-17
temporary changes for making ET non private

kjain110:


src/main/resources/entity-types/users/simple_custom_field_details.json5
```diff
@@ -2,7 +2,7 @@
   id: '0b70ae1e-a87a-4621-a2d8-143973821e39',
   name: 'simple_user_custom_field_details',
   root: false,
-  private: true,
+  private: false,
```
kjain110, line 5: This should be private only

src/main/resources/application.yml
```diff
@@ -1,7 +1,7 @@
 mod-fqm-manager:
   query-retention-duration: 3h
   max-query-size: ${MAX_QUERY_SIZE:1250000}
-  bypass-permissions: false # local development ONLY
+  bypass-permissions: true # local development ONLY
```
mweaver-ebsco, line 4: Random note: If you set this at runtime through a system property or env var, you can avoid needing to update this file, thus avoiding the possibility of accidentally committing it 🙂

PR #595, opened 2025-01-21
[MODFQMMGR-642] Use custom field UUIDs for BE custom field names

bvsharp:
## Purpose
[MODFQMMGR-642](https://folio-org.atlassian.net/browse/MODFQMMGR-642): Use custom field UUIDs for BE custom field names. This will make it possible to track changes to custom field names 

+  Add migration script for updated custom field names

### DIFF
--- a/src/main/java/org/folio/fqm/migration/strategies/V13CustomFieldRename.java
+++ b/src/main/java/org/folio/fqm/migration/strategies/V13CustomFieldRename.java
@@ -1,0 +1,126 @@
+package org.folio.fqm.migration.strategies;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.apache.commons.lang3.tuple.Pair;
+import org.folio.fql.service.FqlService;
+import org.folio.fqm.migration.MigratableQueryInformation;
+import org.folio.fqm.migration.MigrationStrategy;
+import org.folio.fqm.migration.MigrationUtils;
+import org.folio.fqm.migration.warnings.Warning;
+import org.folio.spring.FolioExecutionContext;
+import org.jooq.DSLContext;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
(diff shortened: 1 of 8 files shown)


kjain110 (approved):

src/main/java/org/folio/fqm/migration/strategies/V13CustomFieldRename.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static org.folio.fqm.repository.EntityTypeRepository.CUSTOM_FIELD_NAME;
+import static org.folio.fqm.repository.EntityTypeRepository.CUSTOM_FIELD_PREPENDER;
+import static org.folio.fqm.repository.EntityTypeRepository.CUSTOM_FIELD_TYPE;
+import static org.folio.fqm.repository.EntityTypeRepository.SUPPORTED_CUSTOM_FIELD_TYPES;
+import static org.jooq.impl.DSL.field;
+
+/**
+ * Version 13 -> 14, handles custom field renaming.
+ * <p>
```
ncovercash, line 30: Don't forget to close this!

src/main/java/org/folio/fqm/migration/strategies/V13CustomFieldRename.java
```diff
@@ -0,0 +1,126 @@
...
+            if (!USERS_ENTITY_TYPE_ID.equals(query.entityTypeId())) {
+              result.set(key, value); // no-op
+              return;
+            }
+
+            Optional<Pair<String, String>> namePair = getNamePairs(executionContext.getTenantId())
+              .stream()
+              .filter(pair -> pair.getLeft().equals(key))
+              .findFirst();
+            String resultKey = namePair.map(Pair::getRight).orElse(key);
```
ncovercash, line 79: Do we want to remove invalid custom fields? It'd make this messier (having to whitelist non-custom fields) so idk how worthwhile that is

ncovercash (approved):

src/main/java/org/folio/fqm/migration/strategies/V13CustomFieldRename.java
bvsharp: imo we're ok without it, especially considering a user could immediately start creating invalid custom fields again after this migration runs. It also seems like the UI handles the situation pretty well.

PR #603, opened 2025-01-27
[MODFQMMGR-657] Fix exports for organizations entity type

bvsharp:
## Purpose
[MODFQMMGR-657](https://folio-org.atlassian.net/browse/MODFQMMGR-657): The export doesn't work for "Organizations" ET


src/main/resources/entity-types/organizations/simple_organizations.json5
```diff
@@ -36,7 +36,7 @@
       dataType: {
         dataType: 'rangedUUIDType',
       },
-      isIdColumn: true,
+      isIdColumn: false,
```
mweaver-ebsco, line 39: Do we need both of these to be ID columns? That seems like it'd just slow things down

If so, I'd add a comment to `_id` to explain why

ncovercash (approved):

src/main/resources/entity-types/organizations/simple_organizations.json5
bvsharp: No, that's what this change is for, to set it to false for `_id`

mweaver-ebsco (approved):

src/main/resources/entity-types/organizations/simple_organizations.json5
mweaver-ebsco: Whoops! I misread

PR #607, opened 2025-01-29
[MODFQMMGR-658] Add required permissions to new entity types

bvsharp:
## Purpose
[MODFQMMGR-658](https://folio-org.atlassian.net/browse/MODFQMMGR-658): Some entity types are still available in the Lists app, even though the user doesn't have any ET permission 

Bonus: remove duplicate drv_purchase_order_details ET


ncovercash (approved):

mweaver-ebsco (approved):

src/main/resources/entity-types/orders/drv_purchase_order_details.json5
mweaver-ebsco, line 1: This makes me so happy :) This was probably going to be the next thing I to do in my current ticket, so no need to bother now :)

PR #611, opened 2025-01-31
Remove redundant speed comparison folder

ncovercash:
We moved this to `tools/speed-comparison`, but the original folder didn't get deleted :(


mweaver-ebsco (approved):
Ahhh... Deleting code just feels soooooo good 😌

PR #612, opened 2025-02-03
[MODFQMMGR-610] Use new entity type join format

ncovercash:
## Purpose
https://github.com/folio-org/folio-query-tool-metadata/pull/82 changes the way entity types are joined; for short, fields are responsible for describing joins to other fields, rather than sources to other sources (for long, read that PR).

With these changes, we need to 1) update the flattening service and 2) update the existing entity types which use cross-entity joins.

## Approach
_mom said it's my turn with the flattening service_

A lot of this is described in comments in the code itself, but the main things that happened here are:
- Splitting `EntityTypeFlatteningService` into three parts:
  - `FromClauseUtils` takes over the `getJoinClause` method from the flattening service, as it was separate from the rest of the code in there (and it's the whole `FROM`, not just `JOIN`s :p)
  - `SourceUtils` does the heavy lifting of injecting aliases/etc from sources
  - And the main recursive flattening stayed in the original service

## ✨ Other changes ✨ 

- Flattened entity types are now cached — we were caching the DB/custom field operations for getting entity types, but this change also caches flattened entity types. This saves a lot of application logic from running each time (flattening the entity type, localization, etc).
- As much of the flattening was pulled out into util methods as possible, making later changes much easier to isolate and test without editing monolithic test cases

## Entity type update status

Entity type updates are still WIP, here's what is/is not done:
- [x] Point all JSONB-extracting `createdByUserId` and `updatedByUserId` to be joinable to `simple_user_details` and `simple_FOLIO_user`
- [x] Update sources in `circulation/composite_loan_details`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `finance/composite_ledger__fund`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `inventory/composite_holdings_record`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `inventory/composite_instances`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `inventory/composite_item_details`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] ❌ Update sources in `invoice/composite_invoice__voucher_line__fund_distribution`
  - [x] Flattens
  - [ ] Queries (+ same data as original) ❌ cannot test against original, as original is broken. This version is still broken due to use of `composite_voucher_line__fund_distribution`
  - [ ] Join clause is the same ❌ cannot test against original, as original is broken. This version is still broken due to use of `composite_voucher_line__fund_distribution`
- [x] ⚠️ Update sources in `invoice/composite_invoice__voucher_line__ledger_fund__organization`
  - [x] Flattens
  - [x] Queries (+ same data as original) ⚠️ cannot test against original, as original is broken. fixed in this PR
  - [x] Join clause is the same ⚠️ cannot test against original, as original is broken. fixed in this PR
- [x] Update sources in `invoice/composite_invoice__voucher_line__organization`
  - [x] Flattens
  - [ ] Queries (+ same data as original) ❌ cannot test against original, as original is broken. This version is still broken due to use of `composite_voucher_line__fund_distribution`
  - [ ] Join clause is the same ❌ cannot test against original, as original is broken. This version is still broken due to use of `composite_voucher_line__fund_distribution`
- [x] ⚠️ Update sources in `invoice/composite_voucher_line__fund_distribution`
  - [x] Flattens
  - [x] Queries ~~(+ same data as original)~~ ⚠️ cannot test against original, as original is broken. fixed in this PR
  - [x] ~~Join clause is the same~~ ⚠️ cannot test against original, as original is broken. fixed in this PR
  - [x] ⚠️ this contains a semi-broken self-referential join that does not work when any entity type consumes it. It was broken before this PR, and this PR will not fix it any more than to make it work for this ET itself.
    - This will likely need to be fixed by additional logic when flattening (to support aliases in here), or by creating a separate entity type to wrap this source.
- [x] Update sources in `invoice/composite_voucher_line__voucher__ledger_fund`
  - [x] Flattens
  - [x] Queries (+ same data as original) ⚠️ cannot test against original, as original is broken. fixed in this PR
  - [x] Join clause is the same ⚠️ cannot test against original, as original is broken. fixed in this PR
- [x] Update sources in `orders/composite_po_instance`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `orders/composite_purchase_order_lines`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Update sources in `users/composite_users_details`
  - [x] Flattens
  - [x] Queries (+ same data as original)
  - [x] Join clause is the same
- [x] Check the following work for hidden join fields (e.g. `composite_instances` `instance.instance_type_id` (used to join `inst_stat`):
  - [x] Is this needed??? yes, for `composite_instances`' `simple_instance` `instance.instance_type_id` to be joinable
  - [x] Sync query
  - [x] Async query
  - [x] Queries with non-essential field still fails as before
  - [x] Flattened from /entity-types/... does not include non-essential as before
- [x] final verify with all ITs/normal initialization

The changes in simples to declare additional `joinTo`s that are not already in use is by no means comprehensive, but I figured it was a good start. We will likely be adding many more `joinsTo` throughout when the custom entity type feature gets farther and/or we identify new cases where we want things to correlate.


src/main/java/org/folio/fqm/exceptionhandler/FqmExceptionHandler.java
```diff
@@ -29,14 +29,14 @@ public ResponseEntity<String> handleGenericExceptions(Exception ex, WebRequest r
...
+    log.error("Request failed with internal exception. URL: {}. Failure reason:", url, exception);
     return new ResponseEntity<>(exception.getError(), exception.getHttpStatus());
   }
 
   @ExceptionHandler(Exception.class)
   public ResponseEntity<String> handleGenericExceptions(Exception exception, ServletWebRequest webRequest) {
     String url = webRequest.getHttpMethod() + " " + webRequest.getRequest().getRequestURI();
-    log.error("Request failed with an unhandled exception. URL: {}. Failure reason : {}", url, exception.getMessage());
+    log.error("Request failed with an unhandled exception. URL: {}. Failure reason:", url, exception);
     return new ResponseEntity<>(getStackTraceAsString(exception), HttpStatus.INTERNAL_SERVER_ERROR);
```
ncovercash, line 40: Tracebacks are nice 🚀

pom.xml
```diff
@@ -69,9 +69,9 @@
       <version>${folio-spring-base.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.folio</groupId>
-      <artifactId>folio-query-tool-metadata</artifactId>
-      <version>3.1.0-SNAPSHOT</version>
+	    <groupId>com.github.folio-org</groupId>
+	    <artifactId>folio-query-tool-metadata</artifactId>
+	    <version>fqtm-11-jitpack-SNAPSHOT</version>
```
ncovercash, line 74: This is temporary, but I wanted sonar to be ok with it. Once ETs are updated I'll do a similar thing in lists so we can get it in rancher

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -1,48 +1,77 @@
...
+      Caffeine.newBuilder().expireAfterWrite(cacheDurationSeconds, java.util.concurrent.TimeUnit.SECONDS).build();
+  }
+
   public EntityType getFlattenedEntityType(UUID entityTypeId, String tenantId) {
-    return getFlattenedEntityType(entityTypeId, null, tenantId);
+    return entityTypeCache.get(
+      new EntityTypeCacheKey(tenantId, entityTypeId, localizationService.getCurrentLocales()),
+      k -> getFlattenedEntityType(k.entityTypeId(), null, k.tenantId())
+    );
   }
```
ncovercash, line 64: Why are locales here? The flattened entity types are localized as part of flattening, and we don't want to serve someone a different locale

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -1,48 +1,77 @@
...
 
-  private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource sourceFromParent, String tenantId) {
-    // TODO: clean up
```
ncovercash, line 42: I hope these changes are enough to warrant removing this comment 😆

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -68,253 +99,84 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
...
+      childColumns.stream()
+    );
 
-    flattenedEntityType.columns(getFilteredColumns(allColumns).toList());
+    flattenedEntityType.columns(filterEcsColumns(allColumns).toList());
     flattenedEntityType.requiredPermissions(new ArrayList<>(finalPermissions));
     return localizationService.localizeEntityType(flattenedEntityType);
   }
 
-  private String injectSourceAliasIntoViewExtractor(String sourceViewExtractor, Map<String, String> renamedAliases) {
```
ncovercash, line 120: All this moved to `SourceUtils`

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -68,253 +99,84 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
...
-            .type(source.getJoin().getType())
-            .condition(source.getJoin().getCondition())
-            .joinTo(aliasPrefix + source.getJoin().getJoinTo()); // joinTo in subSource was done in the recursive call, but without the prefix, so we need to add it here
-          newSource.join(newJoin);
-        }
-        return newSource;
-      });
-  }
-
-  public String getJoinClause(EntityType flattenedEntityType, String tenantId) {
```
ncovercash, line 151: This moved to `FromClauseUtils`

src/main/java/org/folio/fqm/utils/flattening/FromClauseUtils.java
```diff
@@ -0,0 +1,299 @@
...
+      parentSource = sourceMap.get(parentSource.getJoinedViaEntityType());
+    }
+
+    if (parentSource.getSourceField() == null) {
+      // parent source is not joined, so we do not need to join either
+      return source;
+    }
+
+    return source.join(computeJoin(flattenedEntityType, parentSource));
+  }
```
ncovercash, line 162: Here's where the main magic happens — we examine each database source and determine if it should be joined.

If it's already got a `join` property with a `joinTo`/etc (as in our previous DB-to-DB joins), we don't do anything but substitute the aliases (since it's already there).

However, if it's a child of entity type source(s), it'll traverse the hierarchy until it finds a parent of itself that is joined, and resolve from there.

This allows for cases such as this (and just about any other nesting condition you can think of):

```
db1 --┐ db1 declares a regular SQL join to db2, this works as normal
db2 <-┘
```
```
simple_1 --┐ simple_1 declares an entity type join to simple_2
└- db1     |   db1 sees its parent joins to simple_2; this (and `computeJoin`) finds the specific field used is in `db2`, and creates the condition there
simple_2 <-┘
└- db2
```
```
simple_1 --┐ simple_1 declares an entity type join to simple_2
├- db1 --┐ |   db1 is joined to db2
└- db2 <-┘ |   db2 is unjoined, but sees its parent joins to simple_2; this (and `computeJoin`) finds the specific field used is in `db3`, and creates the condition there
simple_2 <-┘
└- db3
```

Since each simple or composite entity type has the requirement that exactly one source be unjoined, we can guarantee that resolving an entity type into another with a join will maintain this (as the unjoined of child1 will join to another child, depending on the order).

src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
```diff
@@ -0,0 +1,269 @@
...
+    boolean finalPass
+  ) {
+    getAliasReplacementOrder(renamedAliases)
+      .forEach(alias -> {
+        // only replaces things on the first pass
+        String oldAliasReference = ':' + alias;
+        // we use this to ensure we don't replace prefixes of aliases without the rest of the alias
+        String intermediateAliasReference = ":[%s]".formatted(alias);
+        // we only want to remove the :alias format once we're on the final pass (no more parent sources above this one)
+        String newAliasReference = (finalPass ? "\"%s\"" : ":[%s]").formatted(renamedAliases.get(alias));
```
ncovercash, line 79: This is new! Why is it the way that it is?

Nested composites get their source aliases injected multiple times, once for each level of the hierarchy. This was fine when we just had a single level, but anything more caused issues (e.g. `:user -> :users.user -> :nested.users.user`, but the chain would stop as the first `:user` would be replaced with a DB reference). Now, these intermediate translations map to a `:[...]` format, so `:user -> :[users.user] -> :[nested.users.user]`.

This ensures that we don't prematurely replace parts of aliases (e.g. replacing **:user**s.user after the first iteration), particularly in cases where some sources reuse alias names. Ideally we'd change the sourceAlias format overall and use these unambiguous `:[...]` references, but that seems a bit out of scope for this already large PR. The only real reason I can think of for that is if we want to have a source called `uuid` or `text` or something else that would break our casting, and I really hope we don't have a use case for that 🙃

src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  private static Stream<String> getAliasReplacementOrder(Map<String, String> renamedAliases) {
+    // Sort longer aliases before others, since the map was created in prefix order and we want to use the most recently added aliases first
+    // If we don't do this, then we might replace with "abc" before "abc.def" when handling an alias reference like ":abc.def".
+
+    // We can't simply rely on reversing the `keySet` as there are no guarantees on the order we'll start with.
+    return renamedAliases.keySet().stream().sorted((a, b) -> Integer.compare(b.length(), a.length()));
+  }
```
ncovercash, line 146: Previously, this was inlined as a call to `descendingIterator` on the `keySet`. However, `HashMap`s at least have no guaranteed order (as they are based on `HashSet`), so we can only guarantee uniqueness. This implementation sorts by descending length, so we can always ensure `:users.user` is replaced before `:user`.

src/main/resources/entity-types/configuration/simple_config_data.json5
```diff
@@ -156,6 +170,20 @@
       queryable: true,
       visibleByDefault: false,
       valueGetter: ":sourceAlias.jsonb->'metadata'->>'updatedByUserId'",
+      joinsTo: [
+        {
+          targetId: 'bb058933-cd06-4539-bd3a-6f248ff98ee2', // simple_user_details
```
ncovercash, line 175: I would love for us to be able to replace entity type IDs within definitions like this with just the names of the entity type...but that's just a dream for now...

src/test/java/org/folio/fqm/IntegrationTestBase.java
```diff
@@ -33,6 +34,7 @@
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.greaterThan;
 
+@Disabled
```
ncovercash, line 37: Our integration tests don't particularly appreciate entity types that cannot be serialized/stored into the DB. This `@Disabled` will be removed once the problematic definitions are fixed.

src/test/java/org/folio/fqm/service/EntityTypeFlatteningServiceTest.java
```diff
@@ -1,1075 +1,799 @@
...
+      // what is that composite_composite amalgamation?
+      // composite_composite
+      // ├--- composite_wrapper_to_wrapper
+      // |    ├--- composite_1 <-----------┐ <------------------------┐ via composite_1 to simple_2 (D)
+      // |    |    └--- simple_1 -> db (A) | via simple_2 to simple_1 | resolves. simple_1 (A) to simple_2 (D)
+      // |    └--- composite_2 ------------┘                          |
+      // |         └--- simple_2 -> db (B)                            |
+      // └--- composite_1_to_2 ---------------------------------------┘
+      //      ├--- simple_1 -> db (C) ---┐
+      //      └--- simple_2 -> db (D) <--┘
```
ncovercash, line 355: manmade horrors beyond human comprehension...but it's certainly good as a test case

src/test/java/org/folio/fqm/service/EntityTypeFlatteningServiceTest.java
```diff
@@ -1,1075 +1,799 @@
...
+
+    assertThat(actual.getRequiredPermissions(), contains(SIMPLE_1.getRequiredPermissions().toArray()));
+    assertThat(actual.getColumns(), hasSize(6));
+    for (EntityTypeColumn column : actual.getColumns()) {
+      assertThat(column.getName(), startsWith("simple_1."));
+      assertThat(column.getOriginalEntityTypeId(), is(UUID.fromString(SIMPLE_1.getId())));
+      assertThat(column.getValueGetter(), stringContainsInOrder("\"simple_1.source1\""));
+      assertThat(column.getFilterValueGetter(), is(anyOf(nullValue(), stringContainsInOrder("\"simple_1.source1\""))));
+      assertThat(column.getValueFunction(), is(anyOf(nullValue(), stringContainsInOrder("\"simple_1.source1\""))));
+    }
```
ncovercash, line 573: I kept some of the large instance-equality cases from before, but for most of the new ones I took this approach instead, testing the properties we care about — much easier to decipher the test failures, at least.

src/test/java/org/folio/fqm/testutil/TestDataFixture.java
```diff
@@ -144,21 +144,6 @@ public static EntityType getEntityDefinition() {
...
-    List<EntityTypeDefaultSort> sorts,
-    List<UUID> subEntityTypes
-  ) {
-    return new EntityType()
-      .id(UUID.randomUUID().toString())
-      .name(name)
-      .labelAlias(name)
-      .columns(columns)
-      .defaultSort(sorts)
-      .subEntityTypeIds(subEntityTypes);
```
ncovercash, line 159: This was the only place `subEntityTypeIds` has ever been used

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
bvsharp: Feels good to finally see that removed!

bvsharp (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -68,253 +99,84 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
...
-      }
-      else {
-        UUID sourceEntityTypeId = UUID.fromString(source.getId());
-        // Recursively flatten the source and add it to the flattened entity type
-        EntityType flattenedSourceDefinition = getFlattenedEntityType(sourceEntityTypeId, source, tenantId);
+      EntityTypeSource newSource = SourceUtils.copySource(sourceFromParent, source, renamedAliases);
+      flattenedEntityType.addSourcesItem(newSource);
+
+      if (source instanceof EntityTypeSourceEntityType sourceEt) {
+        // Recursively flatten the source and add it to the ourselves
```
mweaver-ebsco, line 109: Typo

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -68,253 +99,84 @@ private EntityType getFlattenedEntityType(UUID entityTypeId, EntityTypeSource so
...
-  /**
-   * This method injects the source alias into the column's value getter and filter value getter.
-   * It also recursively injects the source alias into nested object types and array types.
-   *
-   * @param <T>            The type of the column, which must extend the Field interface.
-   * @param column         The column to inject the source alias into.
-   * @param renamedAliases The map of old aliases to new aliases.
-   * @return The column with the injected source alias.
-   */
-private static <T extends Field> T injectSourceAlias(T column, Map<String, String> renamedAliases, String sourceAlias) {
```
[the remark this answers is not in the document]
mweaver-ebsco: Over the last few years, I've been moving more and more away from the package-by-layer camp to package-by-feature (so instead of having `repository`, `service`, `utils`, etc packages/dirs, use `entity-type`, `query`, `fql`, etc (with further nested packages, as appropriate, obviously)). It just feels weird to have `LocalizationService` right next to `EntityTypeFlatteningService` in the project 😕 This PR looks like it kind of introduces a mix of the 2 approaches, with the introduction of the `flattening` package (which I'd probably put under `entity-type` in my ideal world).

I definitely wouldn't bother trying to switch entirely to one approach vs the other as part of this already large PR, but it's something we should consider in the future as a pretty quick and easy refactor.

Nothing actionable here; it just popped into my mind mid-review 🙂

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -324,4 +186,7 @@ private boolean ecsEnabled() {
     int totalRecords = parsedJson.read("totalRecords", Integer.class);
     return totalRecords > 0;
   }
+
+  // translations are included as part of flattening, so we must cache based on locales, too
+  private record EntityTypeCacheKey(String tenantId, UUID entityTypeId, List<Locale> locales) {}
```
mweaver-ebsco, line 191: Solid comment! Short, understandable, and answers a question that would probably come up every time someone looks at this 🙂

src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
```diff
@@ -95,4 +98,42 @@ public static List<String> getDateFields(EntityType entityType) {
...
+   * @example bar -> Pair.of("", "bar")
+   * @example foo.bar.baz -> Pair.of("foo.bar", "baz")
+   */
+  public static Pair<String, String> splitFieldIntoAliasAndField(String field) {
+    int dotIndex = field.lastIndexOf('.');
+    if (dotIndex == -1) {
+      return Pair.of("", field);
+    }
+    return Pair.of(field.substring(0, dotIndex), field.substring(dotIndex + 1));
+  }
```
mweaver-ebsco, line 138: Any idea if custom fields will break this? I'm not sure what exactly we use for the field name when we inject those into an ET. If it's something users can influence at all, splitting on a token like this makes me nervous

src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
```diff
@@ -95,4 +98,42 @@ public static List<String> getDateFields(EntityType entityType) {
...
+   * @example foo.bar -> Pair.of("foo", "bar")
+   * @example bar -> Pair.of("", "bar")
+   * @example foo.bar.baz -> Pair.of("foo.bar", "baz")
+   */
+  public static Pair<String, String> splitFieldIntoAliasAndField(String field) {
+    int dotIndex = field.lastIndexOf('.');
+    if (dotIndex == -1) {
+      return Pair.of("", field);
+    }
+    return Pair.of(field.substring(0, dotIndex), field.substring(dotIndex + 1));
```
mweaver-ebsco, line 137: You could shrink this down to a one-liner pretty easily with `String.split()`, rather than having to roll your own version. All you have to do is reverse the field name, split once on "\\.", then reverse the resulting array and array contents 🙂 

```suggestion
return Optional.ofNullable(StreamSupport.stream(Spliterators.spliterator(Arrays.stream(new StringBuilder("blah.bleh.blih").reverse().toString().split("\\.", 2)).map(StringBuilder::new).map(StringBuilder::reverse).collect(Collectors.toCollection(LinkedList::new)).descendingIterator(), 2, Spliterator.ORDERED), false).toList()).map(p -> Pair.of(p.get(0), p.get(1))).orElse(null);
```

Simple!

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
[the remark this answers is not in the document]
ncovercash: agreed — I'd be happy to move it around, I feel the same (for migration too)...but later

src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
ncovercash: It gets put in at the end, and thanks to Bobby's work they're all UUIDs `_custom_field_2c4e9797-422f-4962-a302-174af09b23f9`, so there's nothing to split.

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
ncovercash: at least one person is reading the comments 😆

src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
mweaver-ebsco: Note: I'm pretty sure that call was on a `LinkedHashMap`, which does guarantee the order :) (I don't think `HashMap.keySet()` returns something that provides `decendingIterator()`)

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
ncovercash: stop it, get some help gif

mweaver-ebsco (approved):

PR #616, opened 2025-02-04
MODFQMMGR-644 Replace the DB sources in composite_po_instance with ET sources

mweaver-ebsco:


src/main/resources/entity-types/orders/composite_po_instance.json5
```diff
@@ -13,363 +13,110 @@
...
+//    {
+//      alias: 'location',
+//      type: 'entity-type',
+//      id: '74ddf1a6-19e0-4d63-baf0-cd2da9a46ca4', // simple_locations
+//      join: {
+//        type: 'LEFT JOIN',
+//        joinTo: 'pol.pol',
+//        condition: "jsonb_extract_path_text(:that.jsonb, 'locations', 'locationId')::uuid = :this.id",
+//      },
+//    },
```
mweaver-ebsco, line 40: These commented out ones weren't actually used for anything when they were DB sources. The commented out version uses an entity type, but I didn't want to introduce entire extra ETs full of fields if they weren't basically there already, so I commented it out to make it easy if we want to add em later.

src/main/resources/entity-types/orders/composite_po_instance.json5
```diff
@@ -13,363 +13,110 @@
...
+      id: '30a5cfad-1868-4f46-86b9-a6ef67e2d9bf', // simple_holdings_records
+      essentialOnly: true,
+      // Because there could theoretically be multiple holdings associated with a single POL ID, we need to include that ID to avoid ID collisions
+      useIdColumns: true,
       join: {
         type: 'LEFT JOIN',
-        joinTo: 'pol',
-        condition: "jsonb_extract_path_text(:that.jsonb, 'locations', 'holdingId')::uuid = :this.id",
+        joinTo: 'pol.pol',
+        condition: "EXISTS (SELECT 1 FROM jsonb_array_elements(:that.jsonb -> 'locations') AS hLoc WHERE (hLoc ->> 'holdingId')::uuid = :this.id)",
```
mweaver-ebsco, line 51: 💀
This join condition makes this soooooooo slow for queries that use holdings fields, since it makes the DB materialize a temp table (with no indexes) which it uses for the join. So in some cases, it might be able to do it faster, if it can materialize that quickly into something pretty small

ncovercash (approved):

kjain110 (approved):

bvsharp (approved):

PR #618, opened 2025-02-05
Fix up permissions on a couple ETs

mweaver-ebsco:


src/main/resources/entity-types/orders/composite_po_instance.json5
```diff
@@ -3,25 +3,23 @@
...
-    'orders.po-lines.item.get',
-    'orders.item.get',
-    'inventory-storage.locations.collection.get',
-    'inventory-storage.holdings.item.get',
-    'inventory-storage.instances.item.get',
-    'users.collection.get',
-    'organizations.organizations.collection.get'
+    // These are inherited from simple_config_data, but are also included here because of the added columns here that also retrieve config data
+    "configuration.entries.collection.get", // Required to search for config entries (e.g., in query conditions)
+    "configuration.entries.item.get" // Required to retrieve specific config entry data (e.g., in query results)
```
bvsharp, line 8: Do you think we need the item-level permission if we already have the collection level one? We have a lot of ETs where we only use the collection-level permission. Not a huge deal either way, but this will need to be added to the system user permissions in mod-lists if we keep this.

kjain110 (approved):

src/main/resources/entity-types/orders/composite_po_instance.json5
mweaver-ebsco: I've been wondering the same thing... I guess it really just depends on what the API provides. If the collection endpoint only returns IDs and not actual config data, then we should definitely require both. If the collection endpoint returns full config entries, then we can probably drop the item one

mweaver-ebsco: ...it returns full config entries. I'll remove the `item` permissions

bvsharp: Testing in rancher, it looks like the collection API provides everything we'd need: 
Screenshot 2025-02-05 at 12 08 26 PM

PR #623, opened 2025-02-11
MODFQMMGR-675. Add Custom Fields support for Order and Order Line entities

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODFQMMGR-675
**Finding order by singleselect value:**
image
**Finding order line by singleselect value:**
image
---------------------------------------------------
**Finding order by radiobutton value:**
image
**Finding order line by radiobutton value:**
image
---------------------------------------------------
**Finding order by textfield value:**
image
**Finding order line by textfield value:**
image
---------------------------------------------------
**Finding order by textarea value:**
image
**Finding order line by textarea value:**
image
---------------------------------------------------


src/main/resources/entity-types/orders/simple_purchase_order_custom_field.json5
```diff
@@ -0,0 +1,356 @@
...
+    {
+      name: 'id',
+      sourceAlias: 'custom_fields',
+      dataType: {
+        dataType: 'rangedUUIDType',
+      },
+      isIdColumn: true,
+      queryable: true,
+      visibleByDefault: false,
+      valueGetter: ':sourceAlias.id',
```
mweaver-ebsco, line 23: To make this more readable, you can remove all of the `sourceAlias` properties and change ":sourceAlias" in the `valueGetter`s to ":custom_fields" like this:

```suggestion
      name: 'id',
      dataType: {
        dataType: 'rangedUUIDType',
      },
      isIdColumn: true,
      queryable: true,
      visibleByDefault: false,
      valueGetter: ':custom_fields.id',
```

🙂

src/main/resources/entity-types/orders/simple_purchase_order_custom_field.json5
```diff
@@ -0,0 +1,356 @@
+{
```
mweaver-ebsco, line 1: Is this entity type necessary?

SerhiiNosko: I will try with usinge simple custom field from users pacakge, because it is the same object, thanks

src/main/resources/db/changelog/changes/v1.1.0/harvested-views.xml
```diff
@@ -59,6 +59,17 @@
     </createView>
   </changeSet>
 
+  <changeSet id="create_view_purchase_order_custom_fields" author="v-snosko@ebsco.com">
```
mweaver-ebsco, line 62: Can you add a `runAlways="true"` property to this `changeSet`? We'll be adding that to all of the source-view-creating `changeSet`s very soon, so I just want to make sure this one doesn't fall through the cracks if it ends up getting merged immediately before or after that change 🙂

SerhiiNosko: added

src/main/resources/entity-types/orders/simple_purchase_order_custom_field.json5
SerhiiNosko: this file is not needed, deleted, thanks

src/main/java/org/folio/fqm/repository/EntityTypeRepository.java
```diff
@@ -227,6 +227,7 @@ private EntityTypeColumn handleBooleanCustomField(String id,
       .valueGetter(String.format(STRING_EXTRACTOR, sourceViewExtractor, refId))
       .labelAlias(name)
       .queryable(true)
+      .essential(true)
```
SerhiiNosko, line 230: It is necessary for custom fields declared in simple entity types to be visible in composite ones. In this case, we can filter in the UI by custom fields from both Order and Order Line entities.

mweaver-ebsco (approved):

PR #626, opened 2025-02-11
MODFQMMGR-543 Add support for ordering ET sources in composites

mweaver-ebsco:


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -140,11 +147,21 @@ private EntityType getFlattenedEntityType(
...
+        // If all of the source's sub-sources are non-entity-type sources, then sort the columns alphabetically
+        // Don't sort if there are any ET sub-sources, since they are already sorted by their labelAliases
+        List<EntityTypeColumn> columns = flattenedSourceDefinition
+          .getColumns();
+        EntityType sourceEntityType = entityTypeRepository
+          .getEntityTypeDefinition(sourceEntityTypeId, tenantId)
+          .orElseThrow(() -> new EntityTypeNotFoundException(entityTypeId));
+        long numETSources = sourceEntityType.getSources().stream().filter(EntityTypeSourceEntityType.class::isInstance).count();
+        if (numETSources == 0) {
+          columns = columns.stream().sorted(nullsLast(comparing(entityTypeColumn -> Objects.requireNonNullElse(entityTypeColumn.getLabelAlias(), ""), String.CASE_INSENSITIVE_ORDER))).toList();
```
mweaver-ebsco, line 159: This could probably be formatted a little better...

src/main/java/org/folio/fqm/resource/EntityTypeController.java
```diff
@@ -27,7 +27,7 @@ public class EntityTypeController implements org.folio.fqm.resource.EntityTypesA
   @EntityTypePermissionsRequired
   @Override
   public ResponseEntity<EntityType> getEntityType(UUID entityTypeId, Boolean includeHidden) {
-    return ResponseEntity.ok(entityTypeService.getEntityTypeDefinition(entityTypeId, Boolean.TRUE.equals(includeHidden), true));
+    return ResponseEntity.ok(entityTypeService.getEntityTypeDefinition(entityTypeId, Boolean.TRUE.equals(includeHidden), false));
```
mweaver-ebsco, line 30: With this change, nothing ever actually sets this to true, so I still need to go through and remove it

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -111,9 +117,10 @@ private EntityType getFlattenedEntityType(
     Set<String> finalPermissions = new HashSet<>(originalEntityType.getRequiredPermissions());
     List<EntityTypeColumn> childColumns = new ArrayList<>();
 
-    for (EntityTypeSource source : originalEntityType.getSources()) {
-      EntityTypeSource newSource = SourceUtils.copySource(sourceFromParent, source, renamedAliases);
-      flattenedEntityType.addSourcesItem(newSource);
+    // Sort the sources, so that the resulting columns end up sorted appropriately
+    Iterable<EntityTypeSource> orderedSources = originalEntityType.getSources().stream().sorted(sourceComparator())::iterator;
```
mweaver-ebsco, line 121: Fun trick that probably shouldn't be done very often: Using `Stream::iterator` as an `Iterable`, so that you can an enhanced for-loop with streams (instead of `.forEach()`, which brings some limitations) :)

bvsharp (approved):

kjain110 (approved):

src/main/resources/entity-types/circulation/composite_loan_details.json5
```diff
@@ -7,7 +7,8 @@
       type: 'entity-type',
       targetId: '00349126-c923-4045-a29f-3b6db4c6bea9', // simple_loan
       useIdColumns: true,
-      essentialOnly: true
+      essentialOnly: true,
+      order: 10,
```
mweaver-ebsco, line 11: I decided to set the orders in increments of 10 for 2 reasons:
1. It provides the ability to shuffle things around and add new sources later on with fewer changes later. It just seems like a useful pattern without any real downside that I see
2. Since this is the first ET with a set order, it doubles as some light documentation. By spacing the orders out, it subtly provides some insight into what does and doesn't work (if you want to space them out later, you don't need to go look up code to see if it works when you've already got an example of it working right off the bat)

PR #628, opened 2025-02-11
MODFQMMGR-668:Finance Group ET part 2

kjain110:


src/main/resources/entity-types/finance/simple_groups.json5
```diff
@@ -0,0 +1,229 @@
+{
+  id: 'a01d982c-2c59-408e-8771-7ce0a07bc1a3',
+  name: 'simple_groups',
```
ncovercash, line 3: nit: i'd prefer if it was `simple_finance_groups`, to disambiguate from user's patron groups

bvsharp, line 3: Can we make the file name and ET singular (`simple_group`)? That will match the rest of the simple entities.

bvsharp (approved):

PR #631, opened 2025-02-17
MODFQMMGR-543 Refactor column sorting within ET sources

mweaver-ebsco:
This change makes it so all columns are sorted in a group within whatever entity type they are defined in. This way, columns within simples are always sorted (vs before, when they were only sorted when imported into a composite), and any columns defined directly in a composite are also sorted as a group that appears before any imported columns


src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -52,30 +52,34 @@ public List<Locale> getCurrentLocales() {
...
+      .toList();
 
-    return entityType;
+    return entityType.columns(localizedColumns);
   }
 
-  void localizeEntityTypeColumn(EntityType entityType, EntityTypeColumn column) {
-    if (column.getLabelAlias() == null) {
+  public EntityTypeColumn localizeEntityTypeColumn(EntityType entityType, EntityTypeColumn column) {
+    var newColumn = column.toBuilder().build();
```
mweaver-ebsco, line 63: The diff on this method looks pretty intense, but it's just returning a new instance of the column, rather than modifying it in-place

bvsharp (approved):

PR #637, opened 2025-02-19
MODFQMMGR-543 Update source sorting to use localized names instead of aliases

mweaver-ebsco:


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -45,13 +46,6 @@ public class EntityTypeFlatteningService {
 
   private final Cache<EntityTypeCacheKey, EntityType> entityTypeCache;
 
-  // Comparator for sorting sources based on their order, then alias properties
-  private static final Comparator<EntityTypeSource> sourceComparator =
-    comparing((EntityTypeSource source) ->
-      source instanceof EntityTypeSourceEntityType etSource
-        ? etSource.getOrder()
-        : Integer.MAX_VALUE) // Right now this only really affects things inherited from ET sources, so just put DB sources last. This doesn't really affect anything, but it reduces the noise when debugging
-      .thenComparing(EntityTypeSource::getAlias);
```
mweaver-ebsco, line 54: What changed, since it's not super obvious in the diff:
Right here, where it does `getAlias()` on the source has been replaced with a `localizationServoce.getEntityTypeLabel(et.getName())` on the entity type corresponding to the source, if the source is an ET source. It still uses the alias for DB sources, since they don't really matter (the impact of this comparator is only on inherited columns, and DB sources don't directly provide any columns).

Loading the ET from the source is a bit more complicated and requires the tenant ID, so I moved it into a method

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -241,6 +235,21 @@ private boolean ecsEnabled() {
     return totalRecords > 0;
   }
 
+  private Comparator<EntityTypeSource> getSourceComparator(String tenantId) {
+    return comparing((EntityTypeSource source) ->
+      source instanceof EntityTypeSourceEntityType etSource
+        ? etSource.getOrder()
+        : Integer.MAX_VALUE) // Right now this only affects things inherited from ET sources, so just put DB sources last. This doesn't really affect anything, but it reduces the noise when debugging
```
ncovercash, line 242: Can this one be an `if` too, the ternary makes this hard(er) to read

ncovercash (approved):

bvsharp (approved):

src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
mweaver-ebsco: I tried that, but the `if` version ends up looking almost exactly the same, just a little more verbose

```java
    return comparing((EntityTypeSource source) -> {
      if (source instanceof EntityTypeSourceEntityType etSource) {
        return etSource.getOrder();
      }
      return Integer.MAX_VALUE;
    })
```
or

PR #639, opened 2025-02-19
MODFQMMGR-543 Use localized source names for source ordering

mweaver-ebsco:


src/main/java/org/folio/fqm/service/EntityTypeFlatteningService.java
```diff
@@ -235,19 +235,12 @@ private boolean ecsEnabled() {
...
         : Integer.MAX_VALUE) // Right now this only affects things inherited from ET sources, so just put DB sources last. This doesn't really affect anything, but it reduces the noise when debugging
-      .thenComparing(source -> {
-        if (source instanceof EntityTypeSourceEntityType etSource) {
-          EntityType et = entityTypeRepository.getEntityTypeDefinition(etSource.getTargetId(), tenantId)
-            .orElseThrow(() -> new EntityTypeNotFoundException(etSource.getTargetId()));
-          return Objects.toString(localizationService.getEntityTypeLabel(et.getName())); // Use Objects.toString(), to handle nulls gracefully
-        }
-        return source.getAlias();
-      });
+      .thenComparing(source -> Objects.toString(localizationService.localizeSourceLabel(entityType, source.getAlias()))); // Use Objects.toString(), to handle nulls gracefully
```
mweaver-ebsco, line 243: I like this soooo much better! It's more correct and way simpler 🙂

src/main/java/org/folio/fqm/service/LocalizationService.java
```diff
@@ -82,13 +82,17 @@ public EntityTypeColumn localizeEntityTypeColumn(EntityType entityType, EntityTy
     return newColumn;
   }
 
+  String localizeSourceLabel(EntityType entityType, String sourceAlias) {
```
mweaver-ebsco, line 85: I was on the fence about taking the ET name as the first param vs the ET object. I ended up going with the ET object, to make this easier to use (with a string, you have to do more digging to figure out what exactly is needed)

bvsharp (approved):

PR #647, opened 2025-02-24
[MODFQMMGR-511] Retrieve dropdown values for Instance - Source field

bvsharp:
## Purpose
[MODFQMMGR-511](https://folio-org.atlassian.net/browse/MODFQMMGR-511): Retrieve dropdown values for Instance - Source field

Also snuck in one small voucher_line change because I didn't want to do a separate PR

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -26,6 +26,7 @@ import org.folio.querytool.domain.dto.EntityType;
 import org.folio.querytool.domain.dto.EntityTypeColumn;
 import org.folio.querytool.domain.dto.Field;
 import org.folio.querytool.domain.dto.SourceColumn;
+import org.folio.querytool.domain.dto.ValueSourceApi;
 import org.folio.querytool.domain.dto.ValueWithLabel;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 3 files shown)


kjain110 (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -195,7 +196,10 @@ private ColumnValues getFieldValuesFromApi(Field field, String searchText, List<
     Set<ValueWithLabel> resultSet = new HashSet<>();
     for (String tenantId : tenantsToQuery) {
       try {
-        String rawJson = crossTenantHttpClient.get(field.getValueSourceApi().getPath(), Map.of("limit", String.valueOf(COLUMN_VALUE_DEFAULT_PAGE_SIZE)), tenantId);
+        ValueSourceApi valueSourceApi = field.getValueSourceApi();
+        Map<String, String> queryParams = Optional.ofNullable(valueSourceApi.getQueryParams())
+          .orElseGet(() -> Map.of("limit", String.valueOf(COLUMN_VALUE_DEFAULT_PAGE_SIZE)));
```
mweaver-ebsco, line 201: Slightly more efficient, since it doesn't involve instantiating an `Optional`:
```suggestion
        Map<String, String> queryParams = Objects.requireNonNullElseGet(valueSourceApi.getQueryParams(),
        () -> Map.of("limit", String.valueOf(COLUMN_VALUE_DEFAULT_PAGE_SIZE)));
```

mweaver-ebsco: I hate the name of that method soooo much... It has "require non-null" in the name, while simultaneously providing a mechanism for dealing with it being null 🤦

mweaver-ebsco (approved):

PR #659, opened 2025-02-27
[MODFQMMGR-700] Handle target fields more intelligently for reused names

ncovercash:


src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
```diff
@@ -230,22 +230,7 @@ private static EntityTypeSourceEntityType copySource(
...
-              return renamedAliases.get(p.getLeft()) + "." + p.getRight();
-            } else {
-              // we're referring to an inner field potentially within our inner entity type(s),
-              // so we append the whole shebang. Appending only p.getRight() could cause us to drop subsource aliases.
-              return renamedAliases.get(source.getAlias()) + "." + source.getTargetField();
-            }
-          })
-          .orElse(null)
-      )
+      .targetField(source.getTargetField())
```
ncovercash, line 233: we used to do a bunch of fancy stuff to resolve based on the source's alias, or its renamed ones as we travel up the tree...but that gets weird when we have duplicate names in the hierarchy. Now, we just keep the original field name and append the source's alias (as that gets fully qualified, anyways) — much nicer!

bvsharp (approved):

PR #674, opened 2025-03-10
[MODFQMMGR-706] Fix failing queries for Purchase orders with titles ET

bvsharp:
## Purpose
[MODFQMMGR-706](https://folio-org.atlassian.net/browse/MODFQMMGR-706): Fix failing queries for Purchase orders with titles ET


src/main/resources/entity-types/orders/composite_po_instance.json5
```diff
@@ -23,7 +23,6 @@
       sourceField: 'pol.purchase_order_id',
       essentialOnly: true,
       order: 30,
-      overrideJoinDirection: 'right',
```
bvsharp, line 26: Because we use the POL UUID as the id column for this ET, doing a right join to PO causes query failures when we have POs with no purchase order lines (since the right join to PO causes us to get a null for POL UUID for POs with no PO lines). Instead, since we want a row for each unique POL UUID, we can just use the default left join here.

kjain110 (approved):

src/main/resources/entity-types/orders/composite_po_instance.json5
mweaver-ebsco: Wouldn't this cause us to filter out POs without any associated POLs? Is that okay?

mweaver-ebsco (approved):

PR #675, opened 2025-03-11
[MODFQMMGR-695] Fix entity type cache bug in ECS environments

bvsharp:
## Purpose
[MODFQMMGR-695](https://folio-org.atlassian.net/browse/MODFQMMGR-695): Fix entity type cache bug in ECS environments

Replacing null tenantIds with `executionContext.getTenantId()` since null tenantId causes problems with caching definitions

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -28,6 +28,7 @@ import org.folio.querytool.domain.dto.Field;
 import org.folio.querytool.domain.dto.SourceColumn;
 import org.folio.querytool.domain.dto.ValueSourceApi;
 import org.folio.querytool.domain.dto.ValueWithLabel;
+import org.folio.spring.FolioExecutionContext;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Service;
 
@@ -64,6 +65,7 @@ public class EntityTypeService {
   private final PermissionsService permissionsService;
   private final CrossTenantQueryService crossTenantQueryService;
   private final LanguageClient languageClient;
+  private final FolioExecutionContext executionContext;
 
   /**
    * Returns the list of all entity types.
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -24,9 +24,9 @@
   <properties>
     <!-- runtime dependencies -->
     <folio-spring-base.version>9.0.0</folio-spring-base.version>
-    <folio-query-tool-metadata.version>3.1.0-SNAPSHOT</folio-query-tool-metadata.version>
+    <folio-query-tool-metadata.version>3.2.0-SNAPSHOT</folio-query-tool-metadata.version>
```
mweaver-ebsco, line 27: Any chance we can go with 3.1.0 instead? Not a huge deal, but it'll make Sunflower patch releases easier if we need to do any of those before we do any Trillium work

mweaver-ebsco: Also, unless something changes last minute, 3.1.0 is what we'll be using the Sunflower FQM release

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -73,7 +75,7 @@ public class EntityTypeService {
   public List<EntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds, boolean includeInaccessible, boolean includeAll) {
     Set<String> userPermissions = permissionsService.getUserPermissions();
     return entityTypeRepository
-      .getEntityTypeDefinitions(entityTypeIds, null)
+      .getEntityTypeDefinitions(entityTypeIds, executionContext.getTenantId())
```
mweaver-ebsco, line 78: Does this change mean we can get rid of the special handling of `null` in `getEntityTypeDefinitions()` (or whever it ends up happening)?

mweaver-ebsco (approved):

kjain110 (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
bvsharp: Yep!

bvsharp: Actually, this causes a ton of problems in a bunch of different places due to flattening service weirdness and integration test weirdness. I don't think it's worth removing in the scope of this story.

PR #688, opened 2025-03-19
MODFQMMGR-725:Handle Empty Strings

kjain110:


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -162,25 +162,36 @@ private static Condition handleEquals(EqualsCondition equalsCondition, EntityTyp
...
+
     String dataType = getFieldDataType(entityType, notEqualsCondition);
     String filterFieldDataType = getFieldForFiltering(notEqualsCondition, entityType).getDataType().getDataType();
+
     if (RANGED_UUID_TYPE.equals(filterFieldDataType) || OPEN_UUID_TYPE.equals(filterFieldDataType)) {
       try {
         UUID uuidValue = UUID.fromString((String) notEqualsCondition.value());
-        return cast(field, UUID.class).ne(cast(uuidValue, UUID.class));
+        return cast(field, UUID.class).ne(cast(uuidValue, UUID.class))
+          .and(nonEmptyStringCheck);
```
bvsharp, line 179: I don't think this will work, because it's trying to compare a UUID to a string.

kjain110: Since we don't even use != for UUID, basically we can remove it

ncovercash (approved):

mweaver-ebsco (approved):

bvsharp (approved):

PR #692, opened 2025-03-27
[MODFQMMGR-728] Ensure tenant_id column is last component of result_id

bvsharp:
## Purpose
[MODFQMMGR-728](https://folio-org.atlassian.net/browse/MODFQMMGR-728) Ensure tenant_id column is last component of result_id

Now that we're ordering ET columns by their localized names, there's a chance that the `tenant_id` field can come before the rest of the result_id (e.g. instance_id for instances, holdings_id for holdings, etc.). While this is fine for FQM and lists, this breaks bulk-edit, because it assumes that the tenant_id will always come last in the response from the `/sortedIds` endpoint. In practice, the only ET that this actually affected was Holdings, because tenant_id naturally comes last for the other ETs when ordering columns alphabetically (in English, at least). 

This PR introduces a kind of silly feeling change: now when we get the result ids for an ET, we do a quick check to make sure tenant_id comes last, and move it to the back of the list if necessary. Now bulk-edit is happy again!

NOTE: this will break contents retrieval for cross-tenant Holdings lists that were created between the merging of https://github.com/folio-org/mod-fqm-manager/pull/639 and the merging of this PR. Imo this isn't a huge deal, given that the change has never been released to production, only affects a single ET and only in ECS environments, and there's a simple fix: just refresh the list.

### DIFF
--- a/src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
+++ b/src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
@@ -11,7 +11,7 @@ import org.jooq.Field;
 import org.jooq.SortField;
 import org.jooq.impl.DSL;
 
-import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 
@@ -33,11 +33,10 @@ public class EntityTypeUtils {
    * @return List of id column names for the entity type
    */
   public static List<String> getIdColumnNames(EntityType entityType) {
-    return (!isEmpty(entityType.getColumns()) ? entityType.getColumns() : Collections.<EntityTypeColumn>emptyList())
-        .stream()
-        .filter(column -> Boolean.TRUE.equals(column.getIsIdColumn()))
-        .map(EntityTypeColumn::getName)
-        .toList();
+    return getIdColumns(entityType)
+      .stream()
+      .map(EntityTypeColumn::getName)
+      .toList();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ncovercash (approved):

kjain110 (approved):

src/main/java/org/folio/fqm/utils/EntityTypeUtils.java
```diff
@@ -136,4 +131,27 @@ public static Pair<String, String> splitFieldIntoAliasAndField(String field) {
...
+
+  private static void orderIdColumns(List<EntityTypeColumn> idColumns) {
+    for (int i = 0; i < idColumns.size(); i++) {
+      if (idColumns.get(i).getName().contains("tenant_id")) {
+        EntityTypeColumn tenantIdColumn = idColumns.remove(i);
+        idColumns.add(tenantIdColumn);
+        break;
+      }
+    }
+  }
```
mweaver-ebsco, line 156: Not saying this is a good idea, or that this code will even compile, but I'd consider doing something like this (plus comments) to make `orderIdColumns()` return a new list instead of modifying the existing one. It's slower, but less likely lead to weird bugs 5 years and a dozen refactors from now.
```suggestion
  private static List<EntityTypeColumn> orderIdColumns(List<EntityTypeColumn> idColumns) {
    return idColumns.stream()
        .sorted((a, b) -> {
            if (a.getName().equals(b.getName()) return 0;
            if (a.getName().equals("tenant_id")) return 1;
            if (b.getName().equals("tenant_id")) return -1;
            return 0;
        })
        .toList();
  }
```

mweaver-ebsco (approved):

PR #701, opened 2025-04-06
[MODFQMMGR-724|752] Create composite entity type for order-invoice reports in Lists app/FQM

Saba-Zedginidze-EPAM:
## Purpose
[[MODFQMMGR-724] POC: Add ability to search order related data from Finance, Invoice, Organization domains](https://folio-org.atlassian.net/browse/MODFQMMGR-724)
[[MODFQMMGR-752] Validation for Order-Invoice queries in the Lists app/FQM](https://folio-org.atlassian.net/browse/MODFQMMGR-752)
[[UXPROD-5202] Create composite entity type for order-invoice reports in Lists app/FQM](https://folio-org.atlassian.net/browse/UXPROD-5202)

Spike created to document the outcome: https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/994934786/Spike+MODFQMMGR-752+Create+composite+entity+type+for+order-invoice+reports+in+Lists+app+FQM

## Approach
- Add new composite entity for Order Invoice Analysis
- Add auxiliary entity type to fetch fund data from the fund distribution array
- Add function alias for jsonb_array_elements

## Pre-Merge Checklist
- [x] Added the JSON5 definition to the `src/main/resources/entity-types` directory?
- [x] Checked that querying fields works correctly and all SQL is valid?


src/main/resources/db/changelog/bootstrap/sql/create-view-drv-fund-data.sql
```diff
@@ -0,0 +1,22 @@
...
+SELECT
+  F.jsonb->>'id' as "fund_id",
+	F.jsonb->>'code' as "fund_code",
+	L.jsonb->>'id' as "ledger_id",
+	L.jsonb->>'code' as "ledger_code",
+	FY.jsonb->>'id' as "fiscal_year_id",
+	FY.jsonb->>'code' as "fiscal_year_code",
+	jsonb_agg(B.jsonb->'name') as "budget_names",
+FROM  ${tenant_id}_mod_fqm_manager.src_finance_fund F
+JOIN  ${tenant_id}_mod_fqm_manager.src_finance_budget B ON B.jsonb->>'fundId' = F.id::text
```
SerhiiNosko, line 13: B.fundId would be more performant because it is foreign key with index

src/main/resources/db/changelog/bootstrap/sql/create-view-drv-fund-data.sql
```diff
@@ -0,0 +1,22 @@
...
+JOIN  ${tenant_id}_mod_fqm_manager.src_finance_ledger L ON L.id = F.ledgerId
+JOIN  ${tenant_id}_mod_fqm_manager.src_finance_fiscal_year FY ON B.jsonb->>'fiscalYearId' = FY.id::text
```
SerhiiNosko, line 15: B.fiscalYearId should be used because it has index

src/main/resources/db/changelog/migrations/s-migrations/changelog-s-migrations.xml
Saba-Zedginidze-EPAM, line 1: Needed to add alias for the jsonb_array_elements to the schema as it is getting prefixed by the schema name during query generation

SerhiiNosko (approved):

BKadirkhodjaev (approved):

src/main/resources/entity-types/invoice/composite_invoice_line_fund_distribution.json5
```diff
@@ -0,0 +1,98 @@
...
+    {
+      type: 'entity-type',
+      alias: 'invoice_line',
+      targetId: '1373f063-8b86-48cc-832d-68b4d0dd44fc', // simple_invoice_line
+      useIdColumns: true,
+    },
+    {
+      type: 'db',
+      alias: 'fund_distribution',
+      target: 'jsonb_array_elements("invoice_line_fund_distribution.invoice_line.invoice_lines".jsonb -> \'fundDistributions\')',
```
mweaver-ebsco, line 16: Is this target right? I'd expect "invoice_line_fund_distribution" to show up somewhere in this composite in order for the "invoice_line_fund_distribution.invoice_line.invoice_lines" reference to work

Saba-Zedginidze-EPAM: This auxiliary entity (`composite_invoice_line_fund_distribution`) was created to normalize the fundDistributions array within invoice_line by transforming each array element into a separate row using jsonb_array_elements.

Attempting to perform this normalization directly in the main composite entity (`composite_order_invoice_analytics`) is not viable due to several limitations:
- Alias resolution: You cannot refer to invoice_line.invoice_lines in the target expression unless invoice_line is a source in the main entity. However, adding it as a source introduces redundancy.
- Improper table context: Even if normalization was performed in the main entity, the result would not be treated as a proper table-like source. This means you wouldn't be able to use fields from within the normalized array (e.g., fundId, expenseClassId) in sourceField or join definitions, because they wouldn't resolve properly.
- Column extraction restrictions: Additionally, composite entities do not support referencing their own columns within the same scope. So even if normalization succeeded, extracting and joining on the desired fields from the array would still not be possible within the same entity.

To solve this, normalization is encapsulated in an auxiliary entity. It includes invoice_line as a source and uses a` CROSS JOIN LATERAL` with `jsonb_array_elements` to turn each fund distribution entry into a separate row. This allows proper column extraction and joining with related entities like fund and expense_class.
While this makes the auxiliary entity dependent on the context of the main one (as seen in the target path), it avoids duplicate sources in the main entity and enables a clean and functional setup. The auxiliary entity isn’t independently reusable, but it serves as a critical piece in achieving the necessary normalization and joins.

src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
```diff
@@ -0,0 +1,92 @@
+{
+  id: 'f3ccbf49-8e3e-4f5c-a60e-04ad80543a4a',
+  name: 'composite_order_invoice_analytics',
```
mweaver-ebsco, line 3: Could you add a `description` property to these new entity types? It's a brand-new property that doesn't do anything yet, so there aren't any examples of it out there yet. I would have been surprised if it were already here 😛. It will eventually be user-facing. With the growing number of entity types, it's often hard to see what each one is and what it provides, so we want to start documenting that to make things easier for users.

src/main/resources/entity-types/invoice/composite_invoice_line_fund_distribution.json5
Saba-Zedginidze-EPAM: I have been experimenting for quite some time to be able to use columns of the same entity in other sources' `sourceField` values, however due to [this code](https://github.com/folio-org/mod-fqm-manager/blob/cfea4c5235bdc2d0a3d5dcaa6c7e2601335f3ab2/src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java#L229) every column that is used as a `sourceField` will have some prefix applied to it.
If we directly `sourceField: "fund_id"` then the actual source field will be replresented as `.fund_id`, for which we will receive an error:
`Failure reason: Column .fund_id could not be found`.
Otherwise, we can hardcode column's name to be `composite_order_invoice_analytics.fund_id`, which would resolve the previous issue, but will result in this: 
`"Cannot invoke \"org.folio.querytool.domain.dto.EntityTypeSource.getAlias()\" because \"source\" is null"`
After further investigation, I found that [here](https://github.com/folio-org/mod-fqm-manager/blob/cfea4c5235bdc2d0a3d5dcaa6c7e2601335f3ab2/src/main/java/org/folio/fqm/utils/flattening/FromClauseUtils.java#L291), during sources ordering, the passed source is null, which I tried to fix with an additional null check, but this would later mess up the sources ordering process and result into an incorrectly generated query.

Frankly, I could not yet figure out correct way to use columns in this scenario without modifying Java code, but introcuding an additional source (auxiliary entity) with its own columns resolved it. If you have some suggestion or idea to better accomodate this problem, would gladly appreciate.

I will add detailed comments to the auxiliary entity as well.

translations/mod-fqm-manager/en.json
```diff
@@ -1638,5 +1638,16 @@
...
+  "entityType.composite_order_invoice_analytics.po": "PO",
+  "entityType.composite_order_invoice_analytics.pol": "PO Line",
+  "entityType.composite_order_invoice_analytics.invoice_line_fund_distribution": "Invoice",
+  "entityType.composite_invoice_line_fund_distribution.invoice_line": "Line",
+  "entityType.composite_order_invoice_analytics.invoice": "Invoice",
+  "entityType.composite_order_invoice_analytics.organization": "Organization",
+  "entityType.composite_order_invoice_analytics.fiscal_year": "Fiscal Year",
+  "entityType.composite_order_invoice_analytics.fund": "Fund",
+  "entityType.composite_order_invoice_analytics.fund_type": "Fund Type",
+  "entityType.composite_order_invoice_analytics.expense_class": "Expense Class"
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Thanks, fixed the names

src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
Saba-Zedginidze-EPAM: Added description field with the purpose of this entity as value

mweaver-ebsco (approved):

PR #707, opened 2025-04-11
[MODFQMMGR-750] Inline certain SQL conditions to improve performance

bvsharp:
[MODFQMMGR-750](https://folio-org.atlassian.net/browse/MODFQMMGR-750):  Inline certain SQL conditions to improve performance

### DIFF
--- a/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
+++ b/src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
@@ -77,8 +77,10 @@ public class FqlToSqlConverterService {
   private static final String RANGED_UUID_TYPE = "rangedUUIDType";
   private static final String STRING_UUID_TYPE = "stringUUIDType";
   private static final String OPEN_UUID_TYPE = "openUUIDType";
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -482,9 +486,43 @@ private static String getFieldDataType(EntityType entityType, FieldCondition<?>
   @SuppressWarnings("unchecked")
   private static <T> org.jooq.Field<T> valueField(T value, FieldCondition<?> condition, EntityType entityType) {
     Field field = getField(condition, entityType);
+    boolean inline = shouldInlineValue(entityType, condition, value);
+
     if (field.getValueFunction() != null) {
-      return DSL.field(field.getValueFunction(), (Class<T>) value.getClass(), param("value", value));
+      if (inline) {
+        return DSL.field(field.getValueFunction(), (Class<T>) value.getClass(), DSL.inline(value));
```
mweaver-ebsco, line 493: To be safe, we should be extra double mega sure that this isn't vulnerable to SQL injection attacks

PR #709, opened 2025-04-15
MODFQMMGR-518 Handle zombie queries 🧟

mweaver-ebsco:
## Purpose
This PR is to handle situations where a query is still marked as in-progress when it's not actually running, due to something like a server crash.

## Approach
When retrieving the FQM query details, if the query is in the in-progress status, it will check running queries in the DB to find the corresponding DB query. If it's not found, it re-checks the FQM query status. If the query is still marked as in-progress, then FQM will update its status to failed, then return the failed query details.

### DIFF
--- a/src/main/java/org/folio/fqm/repository/QueryRepository.java
+++ b/src/main/java/org/folio/fqm/repository/QueryRepository.java
@@ -6,6 +6,7 @@ import org.folio.fqm.domain.Query;
 import org.folio.fqm.domain.QueryStatus;
 import org.folio.querytool.domain.dto.QueryIdentifier;
 import org.jooq.DSLContext;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Repository;
 
@@ -29,6 +30,8 @@ public class QueryRepository {
   private static final int MULTIPLE_FOR_STUCK_QUERIES = 3;
 
   private final DSLContext jooqContext;
+  @Qualifier("readerJooqContext")
+  private final DSLContext readerJooqContext;
 
   public QueryIdentifier saveQuery(Query query) {
     jooqContext.insertInto(table(QUERY_DETAILS_TABLE))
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/fqm/repository/QueryRepository.java
```diff
@@ -51,14 +54,62 @@ public void updateQuery(UUID queryId, QueryStatus queryStatus, OffsetDateTime en
...
+        if (query.status() == QueryStatus.IN_PROGRESS && getQueryPids(queryId).isEmpty()) {
+          // 😳! Oh, no.
+          // But wait! Don't panic yet! This could just be a race condition and the query completed before we checked the running queries.
+          log.warn("Query {} has an in-progress status, but no corresponding running SQL query was found. Double-checking...", queryId);
+          if (getQuery(queryId).filter(q2 -> q2.status() == QueryStatus.IN_PROGRESS).isPresent()) {
+            // Oh, no.
+            log.error("Query {} still has an in-progress status, but no corresponding running SQL query. Marking it as failed", queryId);
+            // This is business logic, so it probably doesn't belong in the repository layer, but there's not really another convenient place to put it
+            updateQuery(queryId, QueryStatus.FAILED, OffsetDateTime.now(), "No corresponding running SQL query was found.");
+            return getQuery(queryId);
```
mweaver-ebsco, line 87: This isn't the most efficient approach, but I figured this is a rare edge case, so 1 extra DB query shouldn't be a big deal

src/main/java/org/folio/fqm/repository/QueryRepository.java
```diff
@@ -51,14 +54,61 @@ public void updateQuery(UUID queryId, QueryStatus queryStatus, OffsetDateTime en
...
   @Cacheable(value = "queryCache", condition = "#useCache==true")
   public Optional<Query> getQuery(UUID queryId, boolean useCache) {
+    return getQuery(queryId)
+      .flatMap(query -> {
+        // If the query is in progress, double-check in the DB to see if there's a running SQL query
+        if (query.status() == QueryStatus.IN_PROGRESS && getQueryPids(queryId).isEmpty()) {
+          // 😳! Oh, no.
+          // But wait! Don't panic yet! This could just be a race condition and the query completed before we checked the running queries.
+          log.warn("Query {} has an in-progress status, but no corresponding running SQL query was found. Double-checking...", queryId);
+          if (getQuery(queryId).filter(q2 -> q2.status() == QueryStatus.IN_PROGRESS).isPresent()) {
```
mweaver-ebsco, line 82: I worry about read replica lag here, or even just DB contention. If the SQL query finishes, but the `UPDATE` to change the query status is blocked for a bit and this 2nd `getQuery()` runs first, the query could get marked as failed incorrectly 🙁

mweaver-ebsco: Since this is an edge case, maybe do a retry for a bit, in case the status update on a successful query is slow?

ncovercash (approved):

bvsharp (approved):

kjain110 (approved):
deadlit

src/main/java/org/folio/fqm/repository/QueryRepository.java
```diff
@@ -51,12 +72,70 @@ public void updateQuery(UUID queryId, QueryStatus queryStatus, OffsetDateTime en
...
+
+  private Query handleZombieQuery(UUID queryId) {
+    log.error("Query {} still has an in-progress status, but no corresponding running SQL query. Marking it as failed", queryId);
+    updateQuery(queryId, QueryStatus.FAILED, OffsetDateTime.now(), "No corresponding running SQL query was found.");
+    return getQuery(queryId);
+  }
+
+  private static class ZombieQueryException extends RuntimeException {
+    public ZombieQueryException() {
+      super("🧟");
```
mweaver-ebsco, line 119: 🧟

ncovercash (approved):

src/main/java/org/folio/fqm/service/QueryManagementService.java
```diff
@@ -142,6 +176,51 @@ public Optional<QueryDetails> getQuery(UUID queryId, boolean includeResults, int
...
+   * Retrieves a Query by its ID, with validation that fails queries with no backing DB query
+   * <p>
+   * This method performs the following steps:
+   * 1. Attempts to retrieve the query from the database.
+   * 2. If the query is found and its status is IN_PROGRESS, it checks for corresponding running SQL queries.
+   * 3. If no running SQL queries are found, it double-checks the query status to account for potential race conditions.
+   * 4. If the query is still IN_PROGRESS but no running SQL query is found, it updates the query status to FAILED.
+   * <p>
+   * Note: This method contains business logic in the repository layer, which might not be ideal
+   * but is currently the most convenient place for this logic.
```
bvsharp, line 189: This can be removed now right?

bvsharp (approved):

PR #710, opened 2025-04-15
[MODORDERS-1269] Removed pol alerts and reporting codes

damien-git:
## Purpose
[MODORDERS-1269](https://folio-org.atlassian.net/browse/MODORDERS-1269) - Remove alerts and reporting codes from order lines, use a single order line schema

## Approach
- Removed alerts and reporting codes
- Renamed "`composite_purchase_order_lines`" "`purchase_order_lines`"
- Renamed "`simple_purchase_order_line`" "`purchase_order_line`"

## Related PRs
- See [mod-orders-storage](https://github.com/folio-org/mod-orders-storage/pull/477)

## Pre-Merge Checklist


ncovercash (approved):

SerhiiNosko (approved):

ncovercash:
> here we already updated to new versions of orders and po-line interfaces?

yes, the MD accepts `12.0` and `13.0`

src/main/resources/entity-types/orders/simple_purchase_order_line.json5
```diff
@@ -163,31 +163,6 @@
...
-        FROM\
-          jsonb_array_elements_text(:sourceAlias.jsonb->'alerts') AS elems\
-      )",
-      filterValueGetter: "(\
-        SELECT\
-          array_agg(lower(elems.value::text))\
-        FROM\
-          jsonb_array_elements_text(:sourceAlias.jsonb->'alerts') AS elems\
-      )",
-    },
```
mweaver-ebsco, line 190: There should be a migration for the removed fields, too (see `V12PurchaseOrderIdFieldRemoval.java` for an example of how to do that), so that existing saved queries that use these fields won't break

damien-git: @mweaver-ebsco In the highly unlikely event of a user using these fields, wouldn't it be better if the queries break, to get the user attention ? (as long as it doesn't break other things)

ncovercash: @damien-git this will break in an extremely non-obvious way: refreshes of existing lists will fail with no specific reasoning, and the query builder will simply not show the erroneous part of the query as though it had never existed. Implementing a migration as above will provide warnings to the user (which are saved in the list description) of what changed, what their original query was, and a fallback (if applicable)

damien-git: @ncovercash I added a migration. I also renamed `composite_purchase_order_lines` and `simple_purchase_order_line`, since there is no composite po line schema anymore. Please check.

SerhiiNosko (approved):

PR #735, opened 2025-05-15
[MODFQMMGR-396] Add tenant_name field to cross-tenant entity types

bvsharp:
## Purpose
[MODFQMMGR-396](https://folio-org.atlassian.net/browse/MODFQMMGR-396): Add tenant_name field to cross-tenant entity types


src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -59,61 +64,96 @@ public List<String> getTenantsToQuery(EntityType entityType, UUID userId) {
...
    * @param entityType Entity type definition
-   * @return           List of tenants to query
+   * @return List of tenants to query
    */
   public List<String> getTenantsToQueryForColumnValues(EntityType entityType) {
     return getTenants(entityType, executionContext.getUserId());
   }
 
-  private List<String> getTenants(EntityType entityType, UUID userId) {
+  public List<Pair<String, String>> getTenantIdsWithNames(EntityType entityType, UUID userId) {
```
bvsharp, line 75: I figured it's simpler logic to just always get the ids and names together, and then the consumer can use whichever they need (or both).

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -59,61 +63,96 @@ public List<String> getTenantsToQuery(EntityType entityType, UUID userId) {
...
    * @param entityType Entity type definition
-   * @return           List of tenants to query
+   * @return List of tenants to query
    */
   public List<String> getTenantsToQueryForColumnValues(EntityType entityType) {
     return getTenants(entityType, executionContext.getUserId());
   }
 
-  private List<String> getTenants(EntityType entityType, UUID userId) {
+  public List<Pair<String, String>> getTenantIdNamePairs(EntityType entityType, UUID userId) {
```
bvsharp, line 74: I figured it's simpler logic to just always get the ids and names together, and then the consumer can use whichever they need (or both).

src/main/java/org/folio/fqm/service/CrossTenantQueryService.java
```diff
@@ -59,61 +63,89 @@ public List<String> getTenantsToQuery(EntityType entityType, UUID userId) {
...
+      .map(Pair::getLeft)
+      .toList();
+  }
+
   @SuppressWarnings("unchecked")
   // JsonPath.parse is returning a plain List without a type parameter, and the TypeRef (vs Class) parameter to JsonPath.read is not supported by the JSON parser
-  private List<Map<String, String>> getUserTenants(String consortiumId, String userId) {
-    String userTenantResponse = ecsClient.get(
+  private List<Map<String, String>> getUserTenants(String consortiumId, String userId, String tenantId) {
+    String userTenantResponse = crossTenantClient.get(
```
bvsharp, line 145: We have to use the CrossTenantHttpClient here, because consortia/user-tenants doesn't contain any entries for member tenants (meaning that if you're in the member tenant, there's no way to get the central tenant name without sending a cross-tenant request to the central tenant).

src/main/resources/db/changelog/bootstrap/harvested-views-trillium.xml
```diff
@@ -0,0 +1,19 @@
...
+  <changeSet id="create_view_consortia_tenant" author="bsharp@ebsco.com" runAlways="true">
+    <preConditions onFail="CONTINUE">
+      <tableExists tableName="tenant" schemaName="${tenant_id}_mod_consortia_keycloak"/>
+    </preConditions>
+    <createView
+      replaceIfExists="true"
+      viewName="src_consortia_tenant">
+      SELECT * FROM ${tenant_id}_mod_consortia_keycloak.tenant
+    </createView>
+  </changeSet>
```
bvsharp, line 16: This is my major concern with the PR. We need the consortia.tenant table to be able to run queries on tenant_name. But if mod-fqm-manager is enabled in a non-ECS environment, and the tenant is later added to a consortium, then this table likely won't exist when this liquibase script is run, meaning we won't get the needed view for the tenant table. 

If we can't guarantee that this table exists when mod-fqm-manager is enabled, then we need some way to either:
  1. notify mod-fqm-manager when a tenant is added to a consortium.
  2. Handle tenant_name queries without needing this table (this could technically be done with some on-the-fly mapping in the ResultSetService, but feels pretty nasty because we'd be modifying results in the java code instead of pulling straight from the SQL).

I'd prefer option 1 but still need to think about how exactly to handle it.

bvsharp: After some more reflection I _think_ this should get taken care of by our ${central_tenant_id} replacement logic, but will verify before merging.

src/main/resources/db/changelog/bootstrap/sql/create-view-src-consortia-tenant.sql
```diff
@@ -0,0 +1,20 @@
...
+        FROM information_schema.tables
+        WHERE table_schema = '${tenant_id}_mod_consortia_keycloak'
+          AND table_name = 'tenant'
+    ) THEN
+        RETURN QUERY
+        SELECT t.id, t.name
+        FROM ${tenant_id}_mod_consortia_keycloak.tenant t;
+    ELSE
+        RETURN;
+    END IF;
```
bvsharp, line 15: This ensures that we can enabled the module without ECS, but will pick up the tenant names as soon as ECS is enabled.

mweaver-ebsco (approved):

PR #737, opened 2025-05-19
[MODFQMMGR-752] Add fund distribution value and amount columns

Saba-Zedginidze-EPAM:
## Purpose
[[MODFQMMGR-752] Add "Fund distribution value" column to "composite_invoice_line_fund_distribution"](https://folio-org.atlassian.net/browse/MODFQMMGR-752)

## Approach
Add new columns for displaying fund distribution value and amount

## Pre-Merge Checklist

### DIFF
--- a/src/main/resources/entity-types/invoice/composite_invoice_line_fund_distribution.json5
+++ b/src/main/resources/entity-types/invoice/composite_invoice_line_fund_distribution.json5
@@ -100,5 +100,20 @@
         },
       ],
     },
+    {
+      name: 'fund_total_value',
+      dataType: {
+        dataType: 'stringType',
+      },
+      valueGetter: "(\
+        CASE WHEN (:fund_distribution.jsonb->>'distributionType') = 'percentage' \
+          THEN (:fund_distribution.jsonb->>'value') || '%' \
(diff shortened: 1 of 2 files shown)


SerhiiNosko (approved):

src/main/resources/entity-types/invoice/composite_invoice_line_fund_distribution.json5
```diff
@@ -100,5 +100,35 @@
...
+    {
+      name: 'fund_total_value',
+      dataType: {
+        dataType: 'stringType',
+      },
+      valueGetter: "(\
+        CASE WHEN (:fund_distribution.jsonb->>'distributionType') = 'percentage' \
+          THEN (:fund_distribution.jsonb->>'value') || '%' \
+          ELSE (:fund_distribution.jsonb->>'value') \
+        END \
```
mweaver-ebsco, line 112: Could we get away with adding a `distribution_type` field (it might even be there already) and remove the `CASE`/`THEN`/`ELSE` logic here?

These aren't queryable fields, so it's not a huge deal, but it'd result in better performance when retrieving query results.

mweaver-ebsco (approved):

bvsharp (approved):

PR #751, opened 2025-06-05
[MODFQMMGR-788] Add support for entity-type-level filters

bvsharp:
## Purpose
[MODFQMMGR-788](https://folio-org.atlassian.net/browse/MODFQMMGR-788): Add support for entity-type-level filters

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -499,7 +499,7 @@
     },
     {
       "id": "source-storage-records",
-      "version": "3.4"
+      "version": "3.3"
     }
   ],
   "optional": [
--- a/src/main/resources/entity-types/srs/simple_srs_record.json5
+++ b/src/main/resources/entity-types/srs/simple_srs_record.json5
@@ -45,6 +45,16 @@
       valueGetter: ':record_lb.snapshot_id',
     },
     {
+      name: 'matched_id',
+      dataType: {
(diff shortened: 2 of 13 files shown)


src/main/resources/entity-types/srs/simple_srs_record.json5
```diff
@@ -1,7 +1,7 @@
 {
   id: 'd5449a05-fd57-45e5-9383-ea615246f6f9',
   name: 'simple_srs_record',
-  private: true,
+  private: false,
```
bvsharp, line 4: will revert before merge

descriptors/ModuleDescriptor-template.json
```diff
@@ -499,7 +499,7 @@
     },
     {
       "id": "source-storage-records",
-      "version": "3.4"
+      "version": "3.3"
```
bvsharp, line 502: Realized that if this is getting backported to Ramsons, it'll need an earlier version

mweaver-ebsco: Eh, we can worry about that in the CSP release

mweaver-ebsco (approved):

descriptors/ModuleDescriptor-template.json
bvsharp: I guess not, I just figured it'd be easier to have something that's compliant with both releases on master

PR #766, opened 2025-06-16
MODFQMMGR-412 Throw exception on queries with outdated incoming queries

mweaver-ebsco:


src/main/java/org/folio/fqm/service/MigrationService.java
```diff
@@ -64,4 +66,60 @@ public String getVersion(@CheckForNull String fqlQuery) {
...
   }
+
+  /**
+   * Migrates the query information and verifies that only the version has changed.
+   * If anything other than the version changes, throws a MigrationQueryChangedException.
+   *
+   * @param migratableQueryInformation the query information to migrate
+   * @throws MigrationQueryChangedException if anything other than the version changes
+   */
+  public void verifyQueryIsUpToDate(MigratableQueryInformation migratableQueryInformation) {
```
ncovercash, line 77: I'd use a different name to make it clearer what this does (we have other `verify...` methods which check conditions and, if not matching the condition, update the data and return the updated version).

```suggestion
  public void throwExceptionIfQueryOutdated(MigratableQueryInformation migratableQueryInformation) {
```

src/main/java/org/folio/fqm/service/QueryManagementService.java
```diff
@@ -142,6 +151,11 @@ public ResultsetPage runFqlQuery(String query, UUID entityTypeId, List<String> f
         fields.add(idColumn);
       }
     }
+
+    // Verify that the query is up to date before execution
+    MigratableQueryInformation migratableQueryInformation = new MigratableQueryInformation(entityTypeId, query, fields);
+    migrationService.verifyQueryIsUpToDate(migratableQueryInformation);
```
ncovercash, line 157: do we want to do this when getting contents too?

PR #783, opened 2025-06-17
MODFQMMGR-771 Add checks to prevent custom ETs from setting up their own custom fields

mweaver-ebsco:


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -389,11 +389,11 @@ public CustomEntityType createCustomEntityType(CustomEntityType customEntityType
     validateCustomEntityType(null, customEntityType);
     var now = clockService.now();
     var updatedCustomEntityType = customEntityType.toBuilder()
+      .id(customEntityType.getId() == null ? UUID.randomUUID().toString() : customEntityType.getId())
```
mweaver-ebsco, line 392: Sonar thinks this ID can never be null (probably because it's annotated like that) and it doesn't like that, but it totally can, so I'm going to merge anyway

PR #801, opened 2025-06-26
MODFQMMGR-412 Refactor how outdated queries are rejected

mweaver-ebsco:
This commit also refactors how FQM approaches migration versions, to be a bit simpler and easier to work with

### DIFF
--- a/docs/Migration.md
+++ b/docs/Migration.md
@@ -45,7 +45,7 @@ Any change to an entity type that results in a field being removed or renamed sh
 
 1. Create a new migration strategy in `src/main/java/org/folio/fqm/migration/strategies` that extends `AbstractSimpleMigrationStrategy`
    - Your migration strategy should start with `V#`, where `#` is the version being migrated **from**.
-1. Implement `getSourceVersion` and `getTargetMigration` to return the version queries should be at before they are ran through this migration, and the version the query should be at once this migration is complete. Currently, this is the current integer version, and the next ascending integer.
+1. Implement `getMaximumApplicableVersion` to return the last version to which the migration should be applied. Currently, this is the current integer version.
 1. Implement `getLabel` with a developer-friendly description of what is happening; this will be logged and will be useful for debugging any errors. It is recommended to include ticket numbers or other references to make it easy to track down why this migration/change occurred in the first place, too.
 1. Describe the [changes](#changes) in this migration by implementing any of the applicable methods.
 1. Describe the [warnings](#warnings) in this migration by implementing any of the applicable methods.
@@ -53,6 +53,7 @@ Any change to an entity type that results in a field being removed or renamed sh
 1. Update the `CURRENT_VERSION` in `src/main/java/org/folio/fqm/config/MigrationConfiguration.java`.
 1. If something fancy is being done, or you want to go above and beyond, write a custom test. You can see `src/test/java/org/folio/fqm/migration/strategies/TestTemplate` and `V0POCMigrationTest` for a framework that can easily be extended for common test case formats.
    - Implementations of `AbstractSimpleMigrationStrategy` are automatically tested via `MigrationStrategyRepositoryTest`, however, this is just for basic smoke tests and contains no logic to test specifics of an actual migration.
+   - See [Advanced migrations](#advanced-migrations) for more information
 
 Not all use cases can be covered with `AbstractSimpleMigrationStrategy`; in that case, a custom migration will be needed. See [advanced migrations](#advanced-migrations) for more details.
 
@@ -203,8 +204,11 @@ Warnings for more complex use cases include `OperatorBreakingWarning` and `Value
 Advanced migrations that need to do more advanced logic or conditionals, or apply to a range of migration versions, require custom implementations of `MigrationStrategy`. Tutorials for this is outside the scope of this document, however, there are three methods to be implemented here which should unlock full power:
 
 - `getLabel` should return a descriptive label for logging, just like `AbstractSimpleMigrationStrategy`
-- `applies` should return `true` if this migration logic should be run on a given `version`
-- `apply` is responsible for the full transformation of the query, adding warnings if applicable, updating the `_version`, and anything else necessary.
+- `applies` can optionally be implemented to only run migration logic on matching queries
+- `apply` is responsible for the full transformation of the query, adding warnings if applicable, and anything else necessary.
+  - If the migration results in a breaking change, then make sure to set `hadBreakingChanges` to `true` on the
+    `MigrationQueryInformtaion` object returned from `apply`. This prevents queries from running if they are simply not
+    compatible with the running FQM release.
 
 Helper functions are available in `MigrationUtils`; it is highly recommend to build off one of the existing migrations if possible.
 
--- a/src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
+++ b/src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
@@ -1,46 +1,38 @@
 package org.folio.fqm.migration;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.UncheckedIOException;
+import lombok.extern.log4j.Log4j2;
+import org.folio.fql.service.FqlService;
+import org.folio.fqm.config.MigrationConfiguration;
+import org.folio.fqm.migration.warnings.*;
+
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-import lombok.extern.log4j.Log4j2;
-import org.folio.fql.service.FqlService;
-import org.folio.fqm.config.MigrationConfiguration;
-import org.folio.fqm.migration.warnings.EntityTypeWarning;
-import org.folio.fqm.migration.warnings.FieldWarning;
-import org.folio.fqm.migration.warnings.QueryBreakingWarning;
-import org.folio.fqm.migration.warnings.RemovedEntityWarning;
-import org.folio.fqm.migration.warnings.RemovedFieldWarning;
-import org.folio.fqm.migration.warnings.Warning;
 
 @Log4j2
 public abstract class AbstractSimpleMigrationStrategy implements MigrationStrategy {
 
   protected final ObjectMapper objectMapper = new ObjectMapper();
 
-  /** The version migrating FROM */
-  public abstract String getSourceVersion();
-
-  /** The version migrating TO */
-  public abstract String getTargetVersion();
-
-  /** The entity types that got a new UUID */
+  /**
+   * The entity types that got a new UUID
+   */
   public Map<UUID, UUID> getEntityTypeChanges() {
     return Map.of();
   }
 
   /**
    * The fields that were renamed. Keys should use their OLD entity type ID, if applicable.
-   *
+   * <p>
    * The special key "*" on the field map can be used to apply a template to all fields. This value
    * should be a string with a %s placeholder, which will be filled in with the original field name.
    */
@@ -57,10 +49,10 @@ public abstract class AbstractSimpleMigrationStrategy implements MigrationStrate
 
   /**
    * The fields that were deprecated, removed, etc.
-   *
+   * <p>
    * ET keys should use their OLD entity type ID, if applicable.
    * Fields should use their OLD names, if applicable.
-   *
+   * <p>
    * The function will be given the field's name and FQL (if applicable), as a string
    */
   public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getFieldWarnings() {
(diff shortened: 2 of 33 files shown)


src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -68,95 +57,100 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
   @Override
-  public boolean applies(String version) {
-    return this.getSourceVersion().equals(version);
+  public boolean applies(MigratableQueryInformation migratableQueryInformation) {
+    Set<String> fieldChanges = this.getFieldChanges().getOrDefault(migratableQueryInformation.entityTypeId(), Map.of()).keySet();
+    return getEntityTypeChanges().containsKey(migratableQueryInformation.entityTypeId())
+           || fieldChanges.contains("*")
+           || migratableQueryInformation.fields().stream().anyMatch(fieldChanges::contains)
+           || getEntityTypeWarnings().containsKey(migratableQueryInformation.entityTypeId());
   }
```
mweaver-ebsco, line 66: I tried to update `applies()` in all of these strategies to look at the query and see if the migration is needed, now that the version check is done automatically in `MigrationService`

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -68,95 +57,100 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
+      return MigratableQueryInformation
+        .builder()
+        .entityTypeId(MigrationConfiguration.REMOVED_ENTITY_TYPE_ID)
+        .fqlQuery("")
+        .fields(List.of())
+        .warning(getEntityTypeWarnings().get(src.entityTypeId()).apply(src.fqlQuery()))
+        .hadBreakingChanges(true)
+        .build();
+    } else {
+      entityTypeWarning.ifPresent(warnings::add);
```
mweaver-ebsco, line 86: Eh, I don't like this change. The old version wasn't quite as pretty, but it was a lot clearer. I'll revert this.

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -3,13 +3,8 @@
...
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.UncheckedIOException;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
ncovercash, line 6: :(

mweaver-ebsco: The IDE auto-formatter strikes again!

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -68,95 +63,99 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
-                  }
-                }
-                return getNewFieldName(fieldChanges, f);
-              })
-              .filter(f -> f != null)
-              .distinct()
-              .toList()
-          );
+        fql.set(getNewFieldName(fieldChanges, key), value);
+        hadBreakingChanges.set(true);
```
mweaver-ebsco, line 118: Doing this in the lambda that runs for every field isn't exactly performant, but it's also not slow enough for me to really care ¯\\\_(ツ)_/¯

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -68,95 +63,99 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
+      return MigratableQueryInformation
+        .builder()
+        .entityTypeId(MigrationConfiguration.REMOVED_ENTITY_TYPE_ID)
+        .fqlQuery("")
+        .fields(List.of())
+        .warning(getEntityTypeWarnings().get(src.entityTypeId()).apply(src.fqlQuery()))
+        .hadBreakingChanges(true)
+        .build();
+    } else if (entityTypeWarning.isPresent()) {
+      warnings.add(entityTypeWarning.get());
```
ncovercash, line 92: whoah, did we lose a `}`?

src/main/java/org/folio/fqm/migration/MigrationUtils.java
```diff
@@ -212,4 +197,30 @@ private static JsonNode transformIfStringOnly(JsonNode value, UnaryOperator<Stri
   public interface ValueTransformer {
     String apply(String key, String value, Supplier<String> fql);
   }
+
+  private static final Pattern NUMBER_PATTERN = Pattern.compile("^\\d+$");
```
ncovercash, line 201: should go at the top of the class

src/main/java/org/folio/fqm/migration/MigrationUtils.java
```diff
@@ -212,4 +197,30 @@ private static JsonNode transformIfStringOnly(JsonNode value, UnaryOperator<Stri
...
+  public int compareVersions(String version1, String version2) {
```
ncovercash, line 202: can this be static?
```suggestion
  public static int compareVersions(String version1, String version2) {
```

ncovercash (commented):
i have more, just accidentally started review instead of single comme

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -68,95 +63,99 @@ public Map<UUID, Map<String, BiFunction<String, String, FieldWarning>>> getField
...
+        );
+    }
 
-      return result.withWarnings(new ArrayList<>(warnings));
-    } catch (JsonProcessingException e) {
-      log.error("Failed to serialize FQL query", e);
-      throw new UncheckedIOException(e);
+    if (hadBreakingChanges.get()) {
+      result = result.withHadBreakingChanges(true);
     }
```
mweaver-ebsco, line 154: This could also be something like 
```java
result = result.withHadBreakingChanges(hadBreakingChanges.get() | src.hadBreakingChanges());
```
but the readability sucks with that, so I didn't see any real upside

ncovercash (changes requested):
see comments. very much voting to axe `applies` now

src/main/java/org/folio/fqm/migration/MigrationUtils.java
```diff
@@ -212,4 +197,30 @@ private static JsonNode transformIfStringOnly(JsonNode value, UnaryOperator<Stri
...
+      }
+    } else {
+      int result = version1Parts[0].compareTo(version2Parts[0]);
+      if (result != 0) {
+        return result;
+      }
+    }
+    // We know the current version part is equal, so move on to the next set of parts
+    return compareVersions(version1Parts.length == 1 ? null : version1Parts[1], version2Parts.length == 1 ? null : version2Parts[1]);
+  }
```
mweaver-ebsco, line 225: A wild recursive parser appears!

Each iteration through the method only looks at the strings up until the first `.`. If that is a positive integer, then it's treated as an int. Otherwise, it just does simple string comparison. If the current version parts are different, then you get a result. Otherwise, the method calls itself with everything after the first `.`.

Nulls get turned into `0`s so that "1" == "1.0.0.0.0.0" or "1.000000"

src/main/java/org/folio/fqm/migration/strategies/V3RamsonsFieldCleanup.java
```diff
@@ -28,15 +28,10 @@ public String getLabel() {
   }
 
   @Override
```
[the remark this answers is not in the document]
mweaver-ebsco: Wat

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -27,12 +28,6 @@ public abstract class AbstractSimpleMigrationStrategy implements MigrationStrate
 
   protected final ObjectMapper objectMapper = new ObjectMapper();
 
-  /** The version migrating FROM */
-  public abstract String getSourceVersion();
-
-  /** The version migrating TO */
-  public abstract String getTargetVersion();
-
   /** The entity types that got a new UUID */
```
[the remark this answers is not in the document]
mweaver-ebsco: Good call!

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
mweaver-ebsco: Agreed. I like the idea of letting migration determine if they really need to run (especially if there are a lot of them), but it's just asking for bugs to go through and add them all in a single PR like this. I'd be more in favor of just returning `true` for all of the existing `applies()` implementations now and letting people implement it as needed in future ones.

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
mweaver-ebsco: We did! The removal of the version change made the `try` block unnecessary

src/main/java/org/folio/fqm/migration/MigrationUtils.java
mweaver-ebsco: I'm always on the fence about things like this. On the one hand, I like putting constants at the top of the class. On the other hand, it's only ever used in this one method, and it's only a class field because `Pattern` objects are kind of expensive to build. I wish Java had some nice way to scope it to the method but still make it static 🙁

src/main/java/org/folio/fqm/migration/MigrationUtils.java
mweaver-ebsco: Oh! I guess the `@UtilityClass` annotation forces all methods to be `static`, which is why it was behaving statically despite the missing `static`. I had no idea it did that!

src/main/java/org/folio/fqm/migration/strategies/V10OrganizationStatusValueChange.java
```diff
@@ -2,6 +2,8 @@
 
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
```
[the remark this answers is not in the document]
mweaver-ebsco: Whoops, I must have started to add it, then partially backed that change out. Thanks!

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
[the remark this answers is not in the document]
mweaver-ebsco: I think you're right. Instead of just blindly calling it breaking, this should go up a few lines and only consider it breaking for removed fields or other query breaking warnings here

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
ncovercash: what if we add a default method to the interface? then it can be overridden if necessary, but it's not cluttering up migrations otherwise

src/main/java/org/folio/fqm/service/MigrationService.java
```diff
@@ -43,16 +43,33 @@ public boolean isMigrationNeeded(MigratableQueryInformation migratableQueryInfor
   }
 
   public MigratableQueryInformation migrate(MigratableQueryInformation migratableQueryInformation) {
-    while (isMigrationNeeded(migratableQueryInformation)) {
+    if (migratableQueryInformation.version() == null) {
```
mweaver-ebsco, line 46: I changed this from a loop into a one-shot `if` block, since it doesn't seem like we actually need it to loop right now. @ncovercash, can you think of any cases right now where we'd need it to loop?

docs/Migration.md
```diff
@@ -204,7 +205,10 @@ Advanced migrations that need to do more advanced logic or conditionals, or appl
 
 - `getLabel` should return a descriptive label for logging, just like `AbstractSimpleMigrationStrategy`
 - `applies` should return `true` if this migration logic should be run on a given `version`
```
ncovercash, line 207:
```suggestion
- `applies` can be implemented to only run migration logic on matching queries
```

src/main/java/org/folio/fqm/migration/AbstractSimpleMigrationStrategy.java
```diff
@@ -57,106 +49,100 @@ public Map<UUID, Function<String, EntityTypeWarning>> getEntityTypeWarnings() {
...
-
-            warnings.add(warning);
-            if (warning instanceof RemovedFieldWarning || warning instanceof QueryBreakingWarning) {
-              return;
-            }
+    Map<String, String> fieldChanges = this.getFieldChanges().getOrDefault(src.entityTypeId(), Map.of());
+    Map<String, BiFunction<String, String, FieldWarning>> fieldWarnings =
+      this.getFieldWarnings().getOrDefault(src.entityTypeId(), Map.of());
+
+    AtomicBoolean hadBreakingChanges = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
ncovercash: ...copilot, it is being used inside a lambda nine lines later. 🙄

src/main/java/org/folio/fqm/service/MigrationService.java
ncovercash: the looping was for cases where the strategies weren't in order. right now, they're all in order, so it can be removed (the while loop was exiting after one pass anyways).

src/main/java/org/folio/fqm/migration/MigrationUtils.java
```diff
@@ -212,4 +197,59 @@ private static JsonNode transformIfStringOnly(JsonNode value, UnaryOperator<Stri
...
+        return result;
+      }
+    } else {
+      int result = version1Parts[0].compareTo(version2Parts[0]);
+      if (result != 0) {
+        return result;
+      }
+    }
+    // If we got to here, then the current version parts are equal, so move on to the next set of parts
+    return compareVersions(version1Parts.length == 1 ? null : version1Parts[1], version2Parts.length == 1 ? null : version2Parts[1]);
```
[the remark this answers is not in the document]
mweaver-ebsco: I considered this approach (and even implemented it nearly identically so that I could compare the 2 approaches), but opted for recursion because the code is a bit simpler (in that the method only has to deal with a single version piece at a time without needing to loop through the pieces and performance is not a concern here. The structure leaves parts of it slightly unintuitive, which I attempted to address with comments.

PR #815, opened 2025-07-07
[MODFQMMGR-824] Add inventory information to order-invoice analysis (UAT feedback)

Saba-Zedginidze-EPAM:
## Purpose
[[MODFQMMGR-824] Add inventory information to order-invoice analysis (UAT feedback) ](https://folio-org.atlassian.net/browse/MODFQMMGR-824)

## Approach
Add new entity type for instances to the order-invoice analysis entity

### DIFF
--- a/src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
+++ b/src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
@@ -25,6 +25,15 @@
       order: 20,
     },
     {
+      alias: 'instance',
+      type: 'entity-type',
+      targetId: '8fc4a9d2-7ccf-4233-afb8-796911839862', // simple_instance
+      targetField: 'id',
+      sourceField: 'pol.instance_id',
+      essentialOnly: true,
+      order: 30
+    },
+    {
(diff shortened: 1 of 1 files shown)


SerhiiNosko (approved):

src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
```diff
@@ -24,6 +24,15 @@
...
       order: 20,
     },
+    {
+      alias: 'instance',
+      type: 'entity-type',
+      targetId: '8fc4a9d2-7ccf-4233-afb8-796911839862', // simple_instance
+      targetField: 'id',
+      sourceField: 'pol.instance_id',
+      essentialOnly: true,
+      order: 30
```
mweaver-ebsco, line 34: This isn't really worth changing in this PR, but it's worth knowing for the future: I'd probably use `25` or something in cases like this. The reason we tend to do 10, 20, 30, etc is for cases like this, so that you can easily add new sources later without having to change all of the `order`s in order to put them wherever you want 🙂

PR #823, opened 2025-07-09
[MODFQMMGR-823] Make invoice/invoice line not required (UAT feedback)

Saba-Zedginidze-EPAM:
## Purpose
[[MODFQMMGR-823] Make invoice/invoice line not required (UAT feedback)](https://folio-org.atlassian.net/browse/MODFQMMGR-823)

## Approach
Add new columns to store all related fiscal years for purchase order with the ability to filter out orders that contain a certain fiscal year (e.g. current)

### DIFF
--- a/src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
+++ b/src/main/resources/entity-types/orders/composite_order_invoice_analytics.json5
@@ -99,5 +99,36 @@
       sourceField: 'invoice_line_fund_distribution.expense_class_id',
       order: 100
     },
+  ],
+  columns: [
+    {
+      name: 'allFiscalYears',
+      sourceAlias: 'invoice',
+      dataType: {
+        dataType: 'jsonbArrayType',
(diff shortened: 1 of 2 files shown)


bvsharp:
I'm confused by this, I see that you added 2 columns to the entity type, but how does that make invoice/invoice line not required?

SerhiiNosko:
> I'm confused by this, I see that you added 2 columns to the entity type (which looks fine), but how does that make invoice/invoice line not required?

User can add filtering by these fields that will satisfy user requirements, tomorrow I will ask Saba to update story name accordingly

Saba-Zedginidze-EPAM:
@bvsharp the initial suspicion regarding the actual issue wat that invoice was a required entity for the results to be retrieved from the database, while in reality issue was that the query was supposed to filter out all results for a PO if any of its invoices had a specific fiscal year. This was not possible without a separate column for each returned row to contain all fiscal years of the PO itself.

PR #841, opened 2025-07-17
[MODFQMMGR-883] Migrate effective library fields for Items ET

bvsharp:
## Purpose
[MODFQMMGR-883](https://folio-org.atlassian.net/browse/MODFQMMGR-883) Migrate "Effective library - Name" and "Effective library - Code" fields for Items ET

As of right now, the latest migration on Ramsons is V12 (meaning queries end up as version 13) and the latest migration on Sunflower is V13 (meaning queries end up as version 14).

## Plan
- The original V14 and V15 (present only in Trillium and later) get incremented by one each, so V14 becomes V15 and V15 becomes V16
- This new migration (call it bugfix-migration) is inserted as V14 in Sunflower and Trillium
- bugfix-migration is inserted as V13 in Ramsons and upgrades queries to version 13.5
- In Sunflower and Trillium, the original V13 becomes V13.5 and is modified to accept queries with version 13.5

## How it should work
- Tenants on Ramsons: tenant will originally be on version 13 queries. When this is deployed with Ramsons CSP, queries will be incremented to 13.5. When upgraded to Sunflower, they'll be incremented 13.5 -> 14 -> 15. When upgraded to trillium, they'll go 15 -> 16 -> 17.

- Tenants already on Sunflower: tenant will originally be on version 14 queries. When this is deployed with Sunflower CSP, queries will go through bugfix-migration as V14.

### DIFF
--- a/src/main/java/org/folio/fqm/migration/strategies/AbstractLibraryValueChangeMigration.java
+++ b/src/main/java/org/folio/fqm/migration/strategies/AbstractLibraryValueChangeMigration.java
@@ -1,0 +1,89 @@
+package org.folio.fqm.migration.strategies;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+import org.folio.fql.service.FqlService;
+import org.folio.fqm.client.LocationUnitsClient;
+import org.folio.fqm.client.LocationUnitsClient.LibraryLocation;
+import org.folio.fqm.migration.MigratableQueryInformation;
+import org.folio.fqm.migration.MigrationStrategy;
+import org.folio.fqm.migration.MigrationUtils;
+import org.folio.fqm.migration.warnings.ValueBreakingWarning;
+import org.folio.fqm.migration.warnings.Warning;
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/fqm/migration/strategies/AbstractLibraryValueChangeMigration.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+/**
+ * Base class for library value change migrations.
+ * Handles transformation of library ID values to their corresponding names/codes.
+ */
+@Log4j2
+@RequiredArgsConstructor
+public abstract class AbstractLibraryValueChangeMigration implements MigrationStrategy {
```
bvsharp, line 26: I originally just copied the V9 code to V16 but sonar didn't like all the duplication, so instead we have this abstract class.

mweaver-ebsco (approved):

PR #848, opened 2025-07-18
[MODFQMMGR-884] Migrate effective library fields for Items ET (Sunflower backport)

bvsharp:
## Purpose
[MODFQMMGR-884](https://folio-org.atlassian.net/browse/MODFQMMGR-884): Migrate effective library fields for Items ET

### DIFF
--- a/src/main/java/org/folio/fqm/migration/strategies/V13_5CustomFieldRename.java
+++ b/src/main/java/org/folio/fqm/migration/strategies/V13_5CustomFieldRename.java
@@ -36,7 +36,6 @@ import static org.jooq.impl.DSL.field;
 @RequiredArgsConstructor
 public class V13_5CustomFieldRename implements MigrationStrategy {
 
-  public static final String SOURCE_VERSION = "13.5";
   public static final String TARGET_VERSION = "14";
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/fqm/migration/strategies/V13_5CustomFieldRename.java
```diff
@@ -26,17 +26,17 @@
...
  *
  * @see https://folio-org.atlassian.net/browse/MODFQMMGR-642 for the addition of this migration
  */
 @Log4j2
 @RequiredArgsConstructor
-public class V13CustomFieldRename implements MigrationStrategy {
+public class V13_5CustomFieldRename implements MigrationStrategy {
 
-  public static final String SOURCE_VERSION = "13";
+  public static final String SOURCE_VERSION = "13.5";
```
ncovercash, line 39: V12 still produces a V13 with these changes, leading to V12 -> V13 -> 💥 with no path forward.

I'd suggest changing line 57 `applies` implementation to `"13".equals(version) || "13.5".equals(version)` or similar. The `SOURCE_VERSION` here does not do anything besides get used in that method

ncovercash (changes requested):

PR #856, opened 2025-07-22
MODFQMMGR-869:Repeatable fields in implementation

kjain110:


src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -271,6 +271,36 @@ private static boolean isDateCondition(FieldCondition<?> fieldCondition, EntityT
...
+    if (ARRAY_TYPE.equals(dataType)) {
+      List<Condition> conditionList = inCondition
+        .value().stream()
+        .map(val -> {
+          var valueArray = cast(array(valueField(val, inCondition, entityType)), String[].class);
+          return arrayOverlap(cast(field, String[].class), valueArray);
+        })
+        .toList();
+      return or(conditionList);
+    }
```
bvsharp, line 285: I don't think it makes sense to OR together a bunch of array overlap operations. We should be able to just do a single arrayOverlap between the array from the DB and the array of values.

kjain110: @bvsharp Done. Updated :)

bvsharp (approved):

PR #867, opened 2025-07-28
[MODFQMMGR-829] Add validation for custom ETs

bvsharp:
## Purpose
[MODFQMMGR-829](https://folio-org.atlassian.net/browse/MODFQMMGR-829): Add validation for custom ETs

Also includes some extra validation for regular ETs

### DIFF
--- a/src/main/java/org/folio/fqm/service/EntityTypeService.java
+++ b/src/main/java/org/folio/fqm/service/EntityTypeService.java
@@ -27,8 +27,11 @@ import org.folio.fqm.repository.EntityTypeRepository;
 import org.folio.querytool.domain.dto.AvailableJoinsResponse;
 import org.folio.querytool.domain.dto.ColumnValues;
 import org.folio.querytool.domain.dto.CustomEntityType;
+import org.folio.querytool.domain.dto.CustomFieldMetadata;
+import org.folio.querytool.domain.dto.CustomFieldType;
 import org.folio.querytool.domain.dto.EntityType;
 import org.folio.querytool.domain.dto.EntityTypeColumn;
+import org.folio.querytool.domain.dto.EntityTypeSource;
 import org.folio.querytool.domain.dto.EntityTypeSourceEntityType;
 import org.folio.querytool.domain.dto.Field;
 import org.folio.querytool.domain.dto.LabeledValue;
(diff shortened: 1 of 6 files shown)


ncovercash (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -376,7 +379,7 @@ private static String getFieldValue(Map<String, Object> allValues, String fieldN
   CustomEntityType getCustomEntityType(UUID entityTypeId) {
     var customET = entityTypeRepository.getCustomEntityType(entityTypeId);
     if (customET == null || !Boolean.TRUE.equals(customET.getIsCustom())) {
-      throw new EntityTypeNotFoundException(entityTypeId,  "Entity type " + entityTypeId + " could not be found or is not a custom entity type.");
+      throw new EntityTypeNotFoundException(entityTypeId, String.format("Entity type %s could not be found or is not a custom entity type.", entityTypeId));
```
mweaver-ebsco, line 382: Definitely not a deal breaker by any means, but in simple cases like this, I prefer string concatenation ¯\\\_(ツ)_/¯

src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -411,26 +415,32 @@ public CustomEntityType updateCustomEntityType(UUID entityTypeId, CustomEntityTy
...
-    }
-    if (entityTypeId != null && !entityTypeId.toString().equals(customEntityType.getId())) {
-      throw new InvalidEntityTypeDefinitionException("The entity type ID in the request body does not match the entity type ID in the URL", entityTypeId);
+      throw new InvalidEntityTypeDefinitionException("Custom entity types must not contain columns", customEntityType);
     }
-    if (customEntityType.getCustomFieldEntityTypeId() != null && customEntityType.getCustomFieldEntityTypeId().equals(customEntityType.getId())) {
-      throw new InvalidEntityTypeDefinitionException("Custom entity types must not refer to themselves with the customFieldEntityTypeId property", customEntityType);
+    if (customEntityType.getCustomFieldEntityTypeId() != null) {
+      throw new InvalidEntityTypeDefinitionException("Custom field entity type id must not be defined for custom entity types", customEntityType);
     }
```
mweaver-ebsco, line 444: Do we not want to support custom fields?

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/EntityTypeService.java
bvsharp: They'll still be able to support custom fields, this is just to handle the new logic where custom fields are handled by an EntityTypeColumn instead of using customFieldEntityTypeId.

PR #871, opened 2025-07-31
MODFQMMGR-886: Adding Migration POC

kjain110:


src/test/java/org/folio/fqm/migration/strategies/MigrationDemonstration.java
```diff
@@ -0,0 +1,116 @@
...
+import org.folio.fqm.migration.MigratableQueryInformation;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Demonstration class showing how the ContainsAnyToInOperatorMigration works
+ * This is a POC to show the migration approach for mapping old operators to new ones.
+ */
+public class MigrationDemonstration {
```
ncovercash, line 16: I'm not sure what this class is here for. It's not a test and there is no reason for a `main`.

src/test/java/org/folio/fqm/service/FqlToSqlConverterServiceTest.java
```diff
@@ -1068,4 +1068,28 @@ void shouldThrowExceptionForInvalidDateValue() {
...
+    // Both should produce equivalent SQL - array overlap condition
+    Condition expectedCondition = arrayOverlap(
+      cast(field("arrayField"), String[].class), 
+      cast(array("value1", "value2"), String[].class)
+    );
+
+    assertEquals(expectedCondition, legacyCondition, "Legacy $contains_any should produce expected SQL");
+    assertEquals(expectedCondition, migratedCondition, "Migrated $in should produce equivalent SQL");
+    assertEquals(legacyCondition, migratedCondition, "Legacy and migrated queries should produce identical SQL");
+  }
```
ncovercash, line 1094: There is no migration logic in `FqlToSqlConverterService`. This test does not test anything and is not relevant to the changes in this PR.

PR #891, opened 2025-08-13
[MODFQMMGR-915] Fix failing queries for "Instance - Note - Note Type"

bvsharp:
## Purpose
[MODFQMMGR-915](https://folio-org.atlassian.net/browse/MODFQMMGR-915): Fix failing queries for "Instance - Note - Note Type"

## Approach
Copy arrayTypes / objectTypes when injecting source aliases to prevent the versions belonging to entity types already in the cache from being modified

### DIFF
--- a/src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
+++ b/src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
@@ -65,33 +65,7 @@ public class SourceUtils {
     }
   }
 
-  private static <T extends Field> T injectSourceAlias(
-    T column,
-    Map<String, String> renamedAliases,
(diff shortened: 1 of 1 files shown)


ncovercash (approved):

src/main/java/org/folio/fqm/utils/flattening/SourceUtils.java
```diff
@@ -65,14 +65,41 @@ public static <T extends Field> T injectSourceAlias(
     }
   }
 
+  public static ObjectType injectSourceAliasForObjectType(
```
mweaver-ebsco, line 68: `void` -> `literally anything else` = ♥️

PR #894, opened 2025-08-14
MODFQMMGR-913 Implement soft-deletion of entity types

mweaver-ebsco:


src/main/java/org/folio/fqm/service/EntityTypeService.java
```diff
@@ -89,6 +90,7 @@ public List<EntityTypeSummary> getEntityTypeSummary(Set<UUID> entityTypeIds, boo
     Set<String> userPermissions = permissionsService.getUserPermissions();
     return entityTypeRepository
       .getEntityTypeDefinitions(entityTypeIds, executionContext.getTenantId())
+      .filter(Predicate.not(EntityType::getDeleted))
```
ncovercash, line 93: these should all probably be `Boolean.FALSE.equals(...)`

mweaver-ebsco: `!Boolean.TRUE.equals(...)` is probably the way to go, since we want `null` to be handled the same as `false`

ncovercash (approved):

PR #1235, opened 2026-04-10
MODFQMMGR-1104 Hide simple_srs_record id and externalId fields

psmagin:
## Purpose
Hide simple_srs_record id and externalId fields
https://folio-org.atlassian.net/browse/MODFQMMGR-1104

## Approach
Mark fields with `essential: false`

## Pre-Merge Checklist

### DIFF
--- a/src/main/resources/entity-types/srs/simple_srs_record.json5
+++ b/src/main/resources/entity-types/srs/simple_srs_record.json5
@@ -31,6 +31,7 @@
       queryable: true,
       visibleByDefault: false,
       essential: true,
+      hidden: true,
       valueGetter: ':record_lb.id',
     },
     {
@@ -101,6 +102,7 @@
       queryable: true,
       visibleByDefault: false,
       essential: true,
+      hidden: true,
       valueGetter: ':record_lb.external_id',
       sourceAlias: 'record_lb',
       joinsTo: [


SvitlanaKovalova1 (approved):

src/main/resources/entity-types/srs/simple_srs_record.json5
```diff
@@ -30,7 +30,7 @@
       },
       queryable: true,
       visibleByDefault: false,
-      essential: true,
+      essential: false,
```
mweaver-ebsco: Should this be non-essential? If this simple is pulled into any composites, using this field in the join, I then I think it has to be essential (the `essential` property controls if it is inherited into composite entity types; if it's not in the composite, I believe FQM will be unable to find the field definition needed to create the `FROM` clause)

src/main/resources/entity-types/srs/simple_srs_record.json5
```diff
@@ -100,7 +100,7 @@
       },
       queryable: true,
       visibleByDefault: false,
-      essential: true,
```
mweaver-ebsco, line 103: Same thing here. I believe that making this non-essential will make it not work right composites (and this one has a `joinsTo` property).

I could definitely be wrong about this, though, so I'd try deploying this branch and seeing if the SRS-related composites still work

mweaver-ebsco (approved):
Pavlo and I discussed it, and it's been tested, so it should be fine. That said, it sounds like `hidden: true` would probably be the safer change. Worst case, something breaks and it's trivial to fix

src/main/resources/entity-types/srs/simple_srs_record.json5
psmagin: Reverted, and set hidden=true

src/main/resources/entity-types/srs/simple_srs_record.json5
psmagin: Reverted, and set hidden=true

PR #1238, opened 2026-04-14
MODFQMMGR-1103: Fix oa bugs

elsenhans:
https://folio-org.atlassian.net/browse/MODFQMMGR-1103

## Purpose

1. Some **date fields** do not work correctly when used as filters, e.g with operator “equals” -> “03/31/2026” (date selected via UI date function). The date is rendered from the DB in the format “3/31/2026”, which is why date-based comparison functions fail. The problem occures in the follwing lists/fields
- OA checklist item: filters “OA checklist item date created” and “OA checklist item last updated”
- Charge: filters „OA charge creation date“ and „OA charge last updated“
- Correspondence: filter „OA correspondence last updated“
- Funding: Filter „OA funding last updated“
- Publication Request: Filter „OA pub request created date“ und „OA pub request last updated”

2. OA **Journal** list shows **inconsistent behaviour,** error messages

- Filtering with the operators “not equal to” -> string (e.g., “yes”) and “is not empty/null” -> “true” or “false” results in the error message “An error occurred. List contains no items” -> tested valid search strings with the fields “OA journal indexed in DOAJ label”, “OA journal indexed in DOAJ value”, “OA journal OA status label”, “OA journal OA status value”, “OA journal work title”

- Filtering with operators “starts with” -> string (e.g., “Journal”, “yes”) and “contains” -> string (e.g., “Journal”, “yes”) returns the error message “An error occurred. List contains no items” -> tested with valid search strings in the fields “OA journal indexed in DOAJ label”, “OA journal indexed in DOAJ value”, “OA journal OA status label”, “OA journal OA status value”, “OA journal work title”
-> The same issues likely exist in other drop-down lists and would need to be resolved in the same way there

## Approach
1. Date fields: Added cast for timestamps
2. OA Journal shows inconsistent behaviour: Adding missing `isIdColumn` for `title_instance_id` and `title_instance_identifier_id`

#### TODOS and Open Questions

## Learning

## Pre-Merge Checklist


elsenhans:
Thanks a lot!

PR #1253, opened 2026-04-24
[MODFQMMGR-1128] Translate and disambiguate language codes in FQM

bvsharp:
## Purpose
[MODFQMMGR-1128](https://folio-org.atlassian.net/browse/MODFQMMGR-1128): Translate language codes during result retrieval


src/main/resources/entity-types/inventory/simple_instance.json5
```diff
@@ -1069,7 +1069,6 @@
       isIdColumn: false,
       queryable: true,
       visibleByDefault: false,
-      localizeForExports: true,
```
bvsharp, line 1072: unneeded now that we always translate language codes

src/main/java/org/folio/fqm/service/ResultSetService.java
```diff
@@ -280,6 +305,51 @@ private Optional<String> localizeCountryCode(String code) {
...
+        String rawCode = value.toString();
+        return languageDisplayMap.getOrDefault(rawCode, LanguageLocalizationUtils.localizeLanguageCode(rawCode, folioLocale));
+      })
+      .toList();
+  }
+
+  private Map<String, String> getLanguageDisplayMap(Locale folioLocale) {
+    try {
+      String rawJson = languageClient.get(executionContext.getTenantId());
+      List<String> codes = JsonPath.parse(rawJson).read("$.facets.languages.values.*.id");
```
mweaver-ebsco: Should we move this to either the language client or into a new service? Not a huge deal, but it just feels very slightly out of scope in this class

src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
```diff
@@ -0,0 +1,151 @@
...
+        label -> localizedValues.stream()
+          .filter(other -> label.equals(other.localizedValue()))
+          .map(LocalizedLanguageValue::rawValue)
+          .filter(Objects::nonNull)
+          .distinct()
+          .count()
+      ));
+
+    return localizedValues.stream()
+      .collect(Collectors.toMap(
```
mweaver-ebsco, line 61: Since we're worried about duplicates in the previous stream, do we need to worry about that in this one, too?

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/ResultSetService.java
bvsharp: good catch, definitely meant to do that! I've moved all that stuff to the LanguageClient.

src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
bvsharp: Not in this instance. We care about labels map to multiple codes, but not codes that map to multiple labels (I would expect it to break before this point due to a duplicate key, but that shouldn't ever happen anyways).

PR #1261, opened 2026-05-28
Pull request from unit_tests_org_folio_fqm_migration_strategies_impl_batch_47cbc5b7_1_b41d0b into test-mate-integration

FolioTestMate:


bvsharp (approved):

src/test/java/org/folio/fqm/migration/strategies/impl/V13CustomFieldRenameTest.java
```diff
@@ -27,6 +28,12 @@
 import org.junit.jupiter.params.provider.Arguments;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+import java.util.Map;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.folio.fqm.repository.EntityTypeRepository.CUSTOM_FIELD_PREPENDER;
+import static org.mockito.Mockito.when;
```
bvsharp: One small idea for improvement for this tool: it would be nice if it was able to keep import ordering consistent (i.e., `java.util.Map` should go with the other `java.util` imports).

mweaver-ebsco (approved):

src/test/java/org/folio/fqm/migration/strategies/impl/V13CustomFieldRenameTest.java
khandramai: @bvsharp , looks this requires manual intervention - I'll update this manually.

PR #1277, opened 2026-06-09
[MODFQMMGR-1171] Match stripes language code translations

bvsharp:
## Purpose
[MODFQMMGR-1171](https://folio-org.atlassian.net/browse/MODFQMMGR-1171): Match stripes language code translations

## Approach
Add all language code translations from stripes. Prioritize these before falling back to java locale translation. 

Current planned approach is to localize based on the stripes translation for the **user** locale (using accept-language). This will fall back to tenant locale if user locale is not set. This provides some nice advantages:
  1. Matches how inventory displays codes.
  2. Just nicer UI in general (if my UI is in Spanish, I probably want my language code translations in Spanish too, regardless of the tenant locale).

With this change, the `a2Code` java fallback path is pretty much dead code. But removing it ends up affecting a few different paths, not ideal for a bugfix, so I think we can just consider that tech debt for now.


src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
```diff
@@ -75,24 +76,51 @@ public static Map<String, String> getLanguageDisplayMap(Iterable<String> codes,
...
+    return localizeLanguageCode(code, folioLocale, null);
+  }
+
+  public static String localizeLanguageCode(String code, Locale folioLocale, TranslationService translationService) {
+    String translation = getLanguageTranslation(code, translationService);
+    if (StringUtils.isNotEmpty(translation)) {
+      return translation;
+    }
+
     String a2Code = LANGUAGE_METADATA.codeToA2Map().get(code);
```
mweaver-ebsco, line 88: Tiniest of nits: If `name` isn't empty, then `a2Code` isn't needed at all, so it's juuuuuuuuuust barely more efficient to move this down a few lines. But realistically, there's a solid chance the JVM will do that optimization at runtime anyway, so the only real reason to do it would be for readability, which is pretty subjective. Like I said, it's the tiniest of nits 😛

src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
```diff
@@ -75,24 +76,51 @@ public static Map<String, String> getLanguageDisplayMap(Iterable<String> codes,
...
+    }
+    String canonicalCode = LANGUAGE_METADATA.codeToCanonicalCodeMap().get(code);
+    if (StringUtils.isEmpty(canonicalCode)) {
+      return null;
+    }
+
+    String translationKey = LANGUAGE_TRANSLATION_TEMPLATE.formatted(canonicalCode);
+    String translation = translationService.format(translationKey);
+    if (StringUtils.isBlank(translation) || translation.equals(translationKey)) {
+      return null;
```
mweaver-ebsco, line 115: Should this really return null if the translation == the translation key?

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
bvsharp: Agreed, but I think it'll be easiest to just make a tech debt ticket to remove all the a2Code stuff.

src/main/java/org/folio/fqm/utils/LanguageLocalizationUtils.java
bvsharp: Yeah, this essentially just causes it to proceed with the next fallback. We wouldn't want to return the translationKey itself, because then they'd be seeing things like `mod-fqm-manager.languages.dut` instead of the language name.

PR #1278, opened 2026-06-11
[MODFQMMGR-1133] New composite entity type (ET) - Agreements + lines

Saba-Zedginidze-EPAM:
## Purpose
[[MODFQMMGR-1133] New composite entity type (ET) - Agreements + lines](https://folio-org.atlassian.net/browse/MODFQMMGR-1133)

## Approach
- Add new period entity type
- Update existing entities
- Add new composite type
- Update translations

## Pre-Merge Checklist

- [x] Added the JSON5 definition to the `src/main/resources/entity-types` directory?
- [x] Checked that querying fields works correctly and all SQL is valid?


SerhiiNosko (approved):

src/main/resources/entity-types/agreements/simple_agreements_agreement_line.json5
```diff
@@ -72,139 +74,141 @@
...
       name: 'ent_type',
+      sourceAlias: 'ent',
       dataType: { dataType: 'stringType' },
       essential: true,
       queryable: true,
       visibleByDefault: true,
-      valueGetter: ':ent.ent_type',
+      valueGetter: ':sourceAlias.ent_type',
+      values: [
+        { label: 'Internal', value: null },
```
mweaver-ebsco: Does this actually work? I'd expect the values to display correctly, but I'm skeptical that queries for `Internal` will work right if that gets translated to `null` (because FQM will do `"whatever.ent_type" = null` instead of `"whatever.ent_type" is null`)

src/main/resources/entity-types/agreements/simple_agreements_agreement_line.json5
```diff
@@ -72,139 +74,141 @@
...
-        name: 'ent_res_name',
+      name: 'ent_res_name',
+      sourceAlias: 'ent',
       dataType: { dataType: 'stringType' },
+      essential: true,
       queryable: true,
-      visibleByDefault: false,
-      valueGetter: ':res.res_name',
+      visibleByDefault: true,
+      valueGetter: "(CASE WHEN :ent.ent_type IS NULL THEN :res.res_name ELSE :ent.ent_resource_name END)"
```
mweaver-ebsco, line 200: Woah, this might actually be the first time we've got a field that references multiple sources in a single `valueGetter`! Neat

There might be another one somewhere, but I can't think of any...

src/main/resources/entity-types/agreements/simple_agreements_org.json5
```diff
@@ -81,17 +87,23 @@
     },
     {
       name: 'org_orgs_name',
+      sourceAlias: 'sao',
```
mweaver-ebsco, line 90: Is this the right source alias? It doesn't really matter, since the getters use `:org` directly, but `sourceAlias` should probably point at the right source, just to be safe

src/main/resources/entity-types/agreements/simple_agreements_org.json5
```diff
@@ -81,17 +87,23 @@
...
+      filterValueGetter: "(SELECT lower(${tenant_id}_mod_organizations_storage.f_unaccent(organizations.jsonb->>'name')) \
+                           FROM ${tenant_id}_mod_fqm_manager.src_organizations organizations \
+                           WHERE organizations.id::text = :org.org_orgs_uuid)",
+      valueFunction: 'lower(${tenant_id}_mod_organizations_storage.f_unaccent(:value))',
       idColumnName: 'org_orgs_uuid',
       source: {
         entityTypeId: 'a1e1b9b8-1f9f-4a01-b8c7-2c8a8a000004',
         columnName: 'org_orgs_name',
+        name: 'organization'
       },
```
mweaver-ebsco, line 107: It doesn't really hurt anything, but you no longer need the `source` property when there's a `valueSourceApi` property on a field/column.

src/main/resources/entity-types/agreements/simple_agreements_agreement_line.json5
mweaver-ebsco: One easy fix, if indexes don't matter for this field: turn `null` into `"null"` in the `valueGetter` with a `CASE`/`WHEN`/`THEN`, and then use a value of `"null"` for `Internal` in `values` here. That should make it handle nulls gracefully, but it'd probably also make it so FQM can't use any indexes for this particular field (if any exist)

src/main/resources/entity-types/agreements/simple_agreements_org.json5
Saba-Zedginidze-EPAM: This approach is used also in another place which effectively allows UI to display a picker modal for organizations in case the dropdown doesn't contain all values

src/main/resources/entity-types/agreements/simple_agreements_org.json5
Saba-Zedginidze-EPAM: Since we have 2 different db sources in this simple ET, a common sourceAlias was needed to be set pn all columns, otherwise the correct source wasn't picked up when generating queries, resulting in an error

src/main/resources/entity-types/agreements/simple_agreements_agreement_line.json5
bvsharp: We also support the concept of default values. For example, you could use a value-label pair of `{label: 'Internal', value: 'internal'}` and set `defaultValue: 'internal'` for the field. This will make it so that any records that are `null` for that value will be treated like they have the value `internal` when querying.

Saba-Zedginidze-EPAM: Thanks, adding default value resolved the problem, works as expected now

src/main/resources/entity-types/agreements/simple_agreements_org.json5
mweaver-ebsco: Ah, you're right! I missed that the `name` was getting added! In general, you don't need `source` when there's an `valueSourceApi`, but ones with special handling (like this one, showing the organization picker modal) are different.

src/main/resources/entity-types/agreements/simple_agreements_agreement_line.json5
mweaver-ebsco: Thanks @bvsharp! I completely forgot about `defaultValue`!

src/main/resources/entity-types/agreements/simple_agreements_org.json5
mweaver-ebsco: If that's the case, then it sounds like there's a bug in FQM that we need to fix. If that's the case, then let's keep this as-is and create a bug ticket in MODFQMMGR, with the details.

The value of `sourceAlias` should only be getting used for 2 things in FQM:
1. resolving `:sourceAlias` in the `valueGetter` and/or `filterValueGetter` into a concrete alias in the generated SQL (which doesn't apply in this case, since it's using `:org` instead of `:sourceAlias`), and
2. figuring out how to join entity types together (meaning that `sourceAlias` is required for columns that are used in a join, and it should be set to the appropriate source for the column).

There's no reason to use a common sourceAlias across columns in the ET. It won't actively break anything (because it's entirely unused in this case), but it's a bit confusing when the sourceAlias is set to the wrong source.

mweaver-ebsco (approved):

PR #1279, opened 2026-06-11
MODFQMMGR-1174 Inherit non-essential fields in composite_organization

mweaver-ebsco:
This makes composite_organization inherit all (incl non-essential) fields from simple_organization. It also makes the nested fields queryable, and fixes some related translations.


src/main/resources/entity-types/organizations/simple_organization.json5
```diff
@@ -350,18 +350,19 @@
...
                 dataType: 'stringType',
               },
-              queryable: false,
+              queryable: true,
               valueGetter: "( SELECT array_agg(elems.value->>'addressLine2') FROM jsonb_array_elements(:sourceAlias.jsonb->'addresses') AS elems)",
-              filterValueGetter: "( SELECT array_agg(lower(elems.value->>'addressLine2')) FROM jsonb_array_elements(:sourceAlias.jsonb->'addresses') AS elems)",
-              valueFunction: 'lower(:value)',
+              // This field tends to end up with trailing spaces, so we need to trim those off
+              filterValueGetter: "( SELECT array_agg(lower(trim(elems.value->>'addressLine2'))) FROM jsonb_array_elements(:sourceAlias.jsonb->'addresses') AS elems)",
+              valueFunction: 'lower(trim(:value))',
```
mweaver-ebsco, line 357: This was a fun discovery :P The actual values in the DB have trailing spaces, so when you query, nothing matches. And because the query builder (I think that's what's doing it) trims the query condition value, you can't actually get a matching result with this field unless we trim it in the `filterValueGetter` and `valueFunction`

bvsharp (approved):

PR #1287, opened 2026-06-22
[MODFQMMGR-1185] Support dynamic MARC subfield querying in FQM

bvsharp:
## Purpose
[MODFQMMGR-1185](https://folio-org.atlassian.net/browse/MODFQMMGR-1185): Support dynamic MARC subfield querying in FQM

### DIFF
--- a/src/main/java/org/folio/fqm/repository/ResultSetRepository.java
+++ b/src/main/java/org/folio/fqm/repository/ResultSetRepository.java
@@ -7,6 +7,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -15,6 +16,7 @@ import java.util.stream.Collectors;
 
 import tools.jackson.databind.ObjectMapper;
 import org.folio.fql.model.Fql;
+import org.folio.fql.model.FqlCondition;
 import org.folio.fqm.exception.FieldNotFoundException;
 import org.folio.fqm.service.EntityTypeFlatteningService;
 import org.folio.fqm.service.EntityTypeInitializationService;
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/fqm/repository/ResultSetRepository.java
```diff
@@ -245,6 +265,19 @@ private List<Map<String, Object>> recordToMap(EntityType entityType, Result<Reco
     return resultList;
   }
 
+  private EntityType augmentWithReferencedMarcFields(EntityType entityType, List<String> fields, String tenantId) {
+    return MarcFieldFactory.addSyntheticColumns(entityType, fields, tenantId);
+  }
```
mweaver-ebsco: Since this is just a one-liner that passes its args through anyway, why not inline it and remove the `augment...()` method?

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
```diff
@@ -401,6 +411,64 @@ private static Condition handleContains(ContainsCondition containsCondition, Ent
...
+        .map(value -> marcRowComparison(
+          marcQueryContext,
+          "=",
+          marcQueryValueField(value, notInCondition, entityType),
+          false
+        ))
+        .toList());
+    }
+    if (fieldCondition instanceof StartsWithCondition startsWithCondition) {
+      return marcRowPatternComparison(
```
mweaver-ebsco: Now that we're on a more modern version of java, would a switch over the type of `fieldCondition` work nicely here?

```
return switch (fieldCondition) {
    case EqualsCondition equalsCondition -> marcRowComparison(
            marcQueryContext,
            "=",
            marcQueryValueField(equalsCondition.value(), equalsCondition, entityType),
            true
          );
...
```

mweaver-ebsco (approved):

src/main/java/org/folio/fqm/service/FqlToSqlConverterService.java
bvsharp: yeah, good call. I'll update it.

